PR #19, opened 2021-04-02
Return kubeconfig as TypeList

anjmao:
Kubeconfig schema should be of type TypeList so it can be used in kubernetes provider by passing host, certs and client certs. It's not possible to use only raw config with local_file as local_file data source doesn't play nicely with terraform state.

If local_file state is stored in tf state and file is removed locally it will not be created again which is a problem for CI pipelines to validate terraforms plans.


examples/cluster_with_kubernetes_objects/main.tf
```diff
@@ -53,29 +53,24 @@ resource "castai_cluster" "example_cluster" {
...
 ##############
 
-resource "local_file" "kubeconfig" {
-  sensitive_content = castai_cluster.example_cluster.kubeconfig
-  filename          = "${path.module}/.kube/config"
-}
-
 provider "kubernetes" {
-  config_path = local_file.kubeconfig.filename
+  host = castai_cluster.example_cluster.kubeconfig.0.host
```
ernestas2k, line 64: Format the changed files so fields are aligned on '='

ernestas2k (dismissed):

ernestas2k (approved):

PR #26, opened 2021-11-10
Eks policies

matas-cast:


castai/policies/policy_test.go
```diff
@@ -0,0 +1,21 @@
+package policies
+
+import (
+"testing"
+)
+
+func TestPolicies(t *testing.T) {
+	t.Run("IAM policy", func(t *testing.T) {
+		_, err := GetIAMPolicy("testaccount")
+		if err != nil {
```
ernestas2k, line 10: lets not ignore and assert the passed values landed (checking that policy string contains "testaccount" or checking for containment of `arn:aws:ec2:*:{{ .AccountNumber }}:instance/*` works for me). Nothing too exhausting, one two asserts to make sure interpolation went right. Applicable to both test cases.

ernestas2k (approved):

ernestas2k (approved):

PR #27, opened 2021-12-09
GKE IAM example

matas-cast:


examples/gke_cluster/variables.tf
```diff
@@ -0,0 +1,19 @@
+variable "castai_api_token" {
+  default = "<<replace-this-with-your-token>>"
+}
+
+variable "castai_url" {
+  default = "<<replace-this-with-castai-api-url>>"
```
viktorasm, line 6: set default to PROD API endpoint - this will not be subject to change most of the time, or just don't implement as variable at all, leaving default value.

examples/gke_cluster/variables.tf
```diff
@@ -0,0 +1,19 @@
...
+
+variable "project_id" {
+  default = "<<replace-this-with-your-gcp-project-id>>"
+}
+
+variable "cluster_name" {
+  default = "<<replace-this-with-your-cluster-name>>"
+}
+
+variable "custom_role_id" {
```
viktorasm, line 17: not clear what to set here as default, add a comment and/or default value

castai/provider.go
```diff
@@ -34,6 +34,7 @@ func Provider() *schema.Provider {
 		DataSourcesMap: map[string]*schema.Resource{
 			"castai_credentials": dataSourceCastaiCredentials(),
 			"castai_cluster":     dataSourceCastaiCluster(),
+			"castai_gcp_user_policies":  dataSourceGcpPolicies(),
```
viktorasm, line 37: name this `castai_gke_user_policies` - we're building kubernetes provider-specific DS, not cloud-specific DS

castai/policies/gcp/policy.go
```diff
@@ -0,0 +1,25 @@
...
+var (
+	//go:embed iam-policy.json
+	Policy []byte
+)
+
+type pols struct {
+	Policies []string `json:"Policies"`
+}
+
+func GetIAMPolicy() ([]string, error) {
```
viktorasm, line 17: add a small test to make sure that this JSON parsing will not fail  at runtime.

castai/provider.go
matas-cast: good point, renamed gcp -> gke

examples/gke_cluster/variables.tf
matas-cast: good point, removed at all, using prod by default

castai/data_source_castai_gke.go
```diff
@@ -0,0 +1,30 @@
...
+
+import (
+	"context"
+	"github.com/castai/terraform-provider-castai/castai/policies/gke"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+func dataSourceGkePolicies() *schema.Resource {
```
varnastadeus, line 11: `dataSourceGKEPolicies`

matas-cast: fixed

varnastadeus (approved):

PR #28, opened 2021-12-10
add data_source castai_eks_settings & eks iam example

ernestas2k:


examples/eks/versions.tf
```diff
@@ -0,0 +1,12 @@
+terraform {
+  required_providers {
+    castai     = {
+      source  = "castai/castai"
+      version = "0.0.0-local"
```
ernestas2k, line 5: remove the explicit version from the example (we use it only in localdev example) as it isn't going to work for theprovider users. Users will download latest public version from registry.

examples/eks/iam.tf
```diff
@@ -0,0 +1,113 @@
+# IAM user required for CAST.AI
+
+provider "castai" {
+  api_url   = "https://console.cast.ai//"
```
ernestas2k, line 4: remove this url (our provider defaults to public api). By the way, api is located at https://api.cast.ai

ernestas2k:
Looks good! Just few nitpicks. Let's merge this PR and further incorporate it in full eks cluster example 👍

examples/eks/iam.tf
```diff
@@ -0,0 +1,113 @@
...
+
+provider "aws" {
+  region  = var.region
+}
+
+locals {
+  iam_user = "castai-eks-${var.cluster_name}"
+  aws_account_id = data.aws_caller_identity.current.account_id
+  instance_profile = "cast-${substr(var.cluster_name,0,40)}-eks-${substr(var.cluster_id,0,8)}"
+  instance_profile_roles = toset([
```
ernestas2k, line 15: Worth extracting the roles lists into the same datasource

matas-cast: agreed

ernestas2k:
Approved! 🎆

matas-cast (approved):

PR #29, opened 2021-12-13
Introduce eks clusters and provide read-only example

ernestas2k:


castai/sdk/util_cluster.go
```diff
@@ -0,0 +1,13 @@
+package sdk
+
+// Currently, sdk doesn't have generated constants for cluster status and agent status, declaring our own.
+const (
+	ClusterStatusReady    = "ready"
+	ClusterStatusDeleting = "deleting"
+	ClusterStatusDeleted  = "deleted"
+	ClusterStatusArchived = "archived"
+
+	ClusterAgentStatusOnline        = "online"
```
r0kas, line 10: Maybe this should be `connected` As that's how we're showing in Console.

examples/eks_cluster_readonly/main.tf
```diff
@@ -0,0 +1,57 @@
...
+  chart           = "castai-agent"
+  cleanup_on_fail = true
+
+  set {
+    name  = "provider"
+    value = "eks"
+  }
+  set_sensitive {
+    name  = "apiKey"
+    value = data.castai_agent_settings.agent_settings.token
```
r0kas, line 55: Shouldn't this be set by calling backend to register new cluster?
What we should set is a token to call backend initially.

ernestas2k: Currently, api.cast.ai provides the agent token and script before registering any cluster. As agreed, we're moving the token output into `castai_eks_cluster` for convenience and to address ease of supporting cluster-controller in the future.

r0kas (approved):

examples/external_cluster_eks/main.tf
```diff
@@ -0,0 +1,167 @@
...
+  name = module.eks.cluster_id
+}
+
+### VPC
+
+module "vpc" {
+  source  = "terraform-aws-modules/vpc/aws"
+  version = "3.11.0"
+
+  name                 = "your-vpc"
```
matas-cast, line 41: Should this be among variables?

PR #32, opened 2021-12-23
feat: introduce cluster_token

varnastadeus:


castai/resource_castai_eks_cluster.go
```diff
@@ -139,44 +147,55 @@ func resourceCastaiEksClusterRead(ctx context.Context, data *schema.ResourceData
...
+		data.Set(FieldEKSClusterAccountId, *resp.JSON200.Eks.AccountId)
+		data.Set(FieldEKSClusterRegion, *resp.JSON200.Eks.Region)
+		data.Set(FieldEKSClusterName, *resp.JSON200.Eks.ClusterName)
+		data.Set(FieldEKSClusterInstanceProfileArn, *resp.JSON200.Eks.InstanceProfileArn)
 	}
 
-	agentToken, err := retrieveAgentToken(ctx, client)
-	if err != nil {
-		return diag.FromErr(err)
+	if _, ok := data.GetOk(FieldEKSClusterAgentToken); !ok {
```
matas-cast, line 168: I'm probably missing something, but why still retrieve the Agent Token if it's switched out for Cluster token in the examples? When would we need it?

varnastadeus: I kept it to not introduce breaking change, we should remove it with next major version release

castai/resource_castai_eks_cluster.go
```diff
@@ -139,44 +147,55 @@ func resourceCastaiEksClusterRead(ctx context.Context, data *schema.ResourceData
...
+		tkn, err := retrieveAgentToken(ctx, client)
+		if err != nil {
+			return diag.FromErr(err)
+		}
+		data.Set(FieldEKSClusterAgentToken, tkn)
+	}
+
+	// Create token only if missing.
+	if _, ok := data.GetOk(FieldEKSClusterToken); !ok {
+		tkn, err := createClusterToken(ctx, client, data.Id())
```
matas-cast, line 178: Could there be issues and is it correct, to create a token on read action?

varnastadeus: you are right, moved it to create

matas-cast (approved):

PR #34, opened 2022-01-07
Docs

matas-cast:
Documentation for CAST AI provider in required format


varnastadeus (approved):

Makefile
```diff
@@ -16,6 +16,11 @@ generate-sdk:
 	@echo "==> Generating castai sdk client"
 	go generate castai/sdk/generate.go
 
+# The following command also rewrites existing documentation
+generate-docs:
+	@echo "==> Generating castai sdk client"
```
viktorasm, line 21: misleading message (should be docs)

imo I'd just drop this log, I know i just executed `make generate-docs`, it doesn't need to reply "generating docs"

viktorasm (approved):

Makefile
matas-cast: good point - copy paste fail

yeah makes sense

PR #39, opened 2022-02-18
feat: add dns_cluster_ip for eks clusters

anjmao:


matas-cast:
Looks good, only problem this will not appear in https://registry.terraform.io/providers/castai/castai/latest/docs/resources/eks_cluster 
until you add something in `docs/resources/eks_cluster.md`
I think it would be good to keep it in-sync

there is a docs generator in `main.go`. but it overwrites existing documentation and generates some boilerplate, so maybe it would be easier to do it manually

matas-cast (approved):

castai/policies/iam-policy.json
```diff
@@ -40,7 +40,8 @@
         "ec2:CreateSecurityGroup",
         "ec2:CreateKeyPair",
         "ec2:DeleteKeyPair",
-        "ec2:CreateTags"
+        "ec2:CreateTags",
+        "ec2:ImportKeyPair"
```
matas-cast, line 44: why is this needed?

anjmao: For feature use to import key pairs to aws.

PR #40, opened 2022-03-02
Delete CAST AI provisioned nodes parameter and remove old example

matas-cast:


examples/eks_cluster/variables.tf
```diff
@@ -21,11 +21,39 @@ variable "aws_secret_access_key" {
...
-
 }
 
 variable "cluster_name" {
   type        = string
   description = "EKS cluster name in AWS account."
 }
 
+# Addresses with whitelisted access to Kubernetes API
+variable "whitelisted_ips" {
```
varnastadeus, line 32: do we really need this?

castai/resource_castai_eks_cluster.go
```diff
@@ -261,8 +266,16 @@ func resourceCastaiEKSClusterDelete(ctx context.Context, data *schema.ResourceDa
 
 		if clusterResponse.JSON200.CredentialsId != nil && agentStatus != sdk.ClusterAgentStatusDisconnected {
 			log.Printf("[INFO] Disconnecting cluster.")
+			deleteNodes := false
```
viktorasm, line 269: nitpick: all this "get optional bool or default value" logic breaks the flow here, wrapping into helper func would help a lot:

```go
response, err := client.ExternalClusterAPIDisconnectClusterWithResponse(ctx, clusterId, sdk.ExternalClusterAPIDisconnectClusterJSONRequestBody{
	DeleteProvisionedNodes: getOptionalBool(data, FieldEKSClusterDeleteNodesOnDisconnect, false)
})
//...

func getOptionalBool(data, field, defaultValue) {
}
```

examples/eks_cluster/variables.tf
matas-cast: good point, we used to have this when we had public/private examples. will remove

castai/resource_castai_eks_cluster.go
matas-cast: good point, fixed

varnastadeus (approved):

PR #42, opened 2022-03-21
Add ssh public key field for eks clusters

anjmao:


matas-cast:
Could you add ssh_public_key to the eks_cluster example?

viktorasm (approved):

PR #43, opened 2022-04-05
feat(CORE-2251): add gke cluster resource

aldor007:


examples/gke_cluster/version.tf
```diff
@@ -2,6 +2,15 @@ terraform {
   required_providers {
     castai = {
       source  = "castai/castai"
+      version = "0.0.0-local"
```
matas-cast, line 5: remove dev version

castai/resource_castai_gke_cluster.go
```diff
@@ -0,0 +1,280 @@
...
+		return nil
+	}
+
+	log.Printf("[INFO] Getting cluster information.")
+
+	resp, err := client.ExternalClusterAPIGetClusterWithResponse(ctx, data.Id())
+	if err != nil {
+		return diag.FromErr(err)
+	} else if resp.StatusCode() == http.StatusNotFound {
+		log.Printf("[WARN] Removing cluster %s from state because it no longer exists in CAST.AI", data.Id())
```
matas-cast, line 148: correct branding is
CAST.AI -> CAST AI

examples/gke_cluster/variables.tf
```diff
@@ -3,3 +3,59 @@ variable "castai_api_token" {}
 variable "project_id" {}
 variable "cluster_region" {}
 variable "cluster_name" {}
+
+variable "castai_api_url" {
```
matas-cast, line 7: I think this should be removed, by default we use it and only adds confusion for external users

matas-cast (approved):

PR #49, opened 2022-04-13
feat: EKS assume role support

matas-cast:


castai/data_source_castai_eks_cluster_id.go
```diff
@@ -0,0 +1,68 @@
...
+	"context"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
+
+	"github.com/castai/terraform-provider-castai/castai/sdk"
+)
+
+const (
+	EKSClusterIDFieldAccountId = "account_id"
```
aldor007, line 13: I wonder if we can have generic data source for getting cluster id. It can be useful for other cloud providers as well.

examples/eks_cluster_assumerole/main.tf
```diff
@@ -0,0 +1,66 @@
+locals {
+  role_name = "matas-04-08"
```
aldor007, line 2: Maybe something more generic? :)

examples/eks_cluster_assumerole/main.tf
```diff
@@ -0,0 +1,66 @@
...
+  cluster_name               = var.cluster_name
+}
+
+data "castai_eks_user_arn" "castai_user_arn" {
+  cluster_id = data.castai_eks_clusterid.castai_cluster_id.id
+}
+
+
+module "castai-eks-role-iam" {
+  source  = "../../../terraform-castai-eks-role-iam"
```
aldor007, line 39: This path should be changed to terraform registry

castai/data_source_castai_eks_cluster_id.go
matas-cast: It might be useful, but now we don't have any need for that so can be skipped I think

PR #51, opened 2022-04-20
fix(gke): add missing support for GKE zonal clusters

aldor007:
PR for usage in GKE module
https://github.com/castai/terraform-castai-gke-cluster/pull/4

### DIFF
--- a/castai/resource_castai_gke_cluster.go
+++ b/castai/resource_castai_gke_cluster.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"log"
 	"net/http"
+	"strings"
 	"time"
 
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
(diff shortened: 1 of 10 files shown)


castai/resource_castai_gke_cluster.go
```diff
@@ -19,6 +19,7 @@ const (
 	FieldGKEClusterName                    = "name"
 	FieldGKEClusterProjectId               = "project_id"
 	FieldGKEClusterRegion                  = "region"
```
matas-cast, line 21: If location is region or region+zone, then region I believe can be made redundant

anjmao: We can keep only location field, but compute region from location since we need it for api
```
{
    "clusterName": "string",
    "location": "string",
    "projectId": "string",
    "region": "string"
  }
  ```
  
  Also you can check if we really need to send region.

Comment (aldor007) [reply castai/resource_castai_gke_cluster.go]: A region removed from the provider but it is still required by our API

File: examples/gke_cluster_zonal/gke.tf [hunk 2]
```diff
@@ -0,0 +1,42 @@
...
+      disk_size_gb       = 100
+      disk_type          = "pd-standard"
+      image_type         = "COS_CONTAINERD"
+      auto_repair        = true
+      auto_upgrade       = true
+      preemptible        = false
+      initial_node_count = 1
+    },
+  ]
+}
```
anjmao, line 42: Add new lines

anjmao (approved):

PR #53, opened 2022-04-25
chore: add example for EKS with NAT and Loadbalancer

aldor007:


examples/eks_cluster_webshop/lb.tf
```diff
@@ -0,0 +1,9 @@
+module "load_balancer_controller" {
+  source = "git::https://github.com/DNXLabs/terraform-aws-eks-lb-controller.git"
```
r0kas, line 2: How did you choose this module? Doesn't look very popular one.

examples/eks_cluster_webshop/main.tf
```diff
@@ -0,0 +1,99 @@
...
+        },
+        "nodeDownscaler": {
+            "emptyNodes": {
+                "enabled": true
+            }
+        }
+    }
+  EOT
+
+}
```
r0kas, line 99: New lines missing on several files.

examples/eks_cluster_webshop/versions.tf
```diff
@@ -0,0 +1,13 @@
+terraform {
+  required_version = ">= 0.13"
```
r0kas, line 2: Why not >= 1.0

examples/eks_cluster_webshop/lb.tf
nilement: reached out to Marcin as well, but seems like a fairly simple module and there isn't any more popular alternatives

wonder if we could replace source with
https://registry.terraform.io/modules/DNXLabs/eks-lb-controller/aws/latest though?

examples/eks_cluster_webshop/main.tf
aldor007: Added

examples/eks_cluster_webshop/versions.tf
aldor007: updated

examples/eks_cluster_webshop/lb.tf
aldor007: I didn't find anything better than this one. Writing it on my own would be copy-paste of this one.
 Source for module updated.

examples/eks_cluster_webshop/eks.tf
```diff
@@ -0,0 +1,100 @@
+#2. create EKS cluster
+module "eks" {
+  source  = "terraform-aws-modules/eks/aws"
+  version = "18.20.5"
+
+  cluster_name    = var.cluster_name
+  cluster_version = "1.21"
```
r0kas, line 7: 1.21 is quite old already. 1.22 maybe? Not sure if this can break anything.
[Kubernetes version 1.22 prerequisites](https://docs.aws.amazon.com/eks/latest/userguide/update-cluster.html#update-1.22)

r0kas (approved):

PR #59, opened 2022-05-05
fix(eks): do not change SGs after first apply

aldor007:
Update in module https://github.com/castai/terraform-castai-eks-cluster/pull/14

### DIFF
--- a/examples/eks_cluster/output.tf
+++ b/examples/eks_cluster/output.tf
@@ -1,0 +1,4 @@
+output "sgs" {
+  description = "EKS cluster security groups"
+  value = module.castai-eks-cluster.security_groups
+}
--- a/examples/eks_cluster/output.tf
+++ b/examples/eks_cluster/output.tf
@@ -1,4 +1,4 @@
 output "sgs" {
   description = "EKS cluster security groups"
-  value = module.castai-eks-cluster.security_groups
+  value       = module.castai-eks-cluster.security_groups
 }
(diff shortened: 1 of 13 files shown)


examples/eks_cluster_webshop/main.tf
```diff
@@ -64,9 +64,6 @@ module "castai-eks-cluster" {
   aws_instance_profile_arn      = module.castai-eks-role-iam.instance_profile_arn
 
   subnets         = var.subnets
-  # As for now (05.05.2022) you need to fill security groups after cluster onboarding with values from plan.
```
r0kas, line 67: How this example will work with SGs from now on?

examples/eks_cluster/output.tf
```diff
@@ -0,0 +1,4 @@
+output "sgs" {
+  description = "EKS cluster security groups"
+  value = module.castai-eks-cluster.security_groups
+}
```
r0kas, line 4: newline

r0kas (approved):

examples/eks_cluster_webshop/main.tf
aldor007: In that example we didn't use SGs it were added only because of this bug

examples/eks_cluster_webshop/main.tf
```diff
@@ -64,9 +64,6 @@ module "castai-eks-cluster" {
   aws_instance_profile_arn      = module.castai-eks-role-iam.instance_profile_arn
 
   subnets         = var.subnets
-  # As for now (05.05.2022) you need to fill security groups after cluster onboarding with values from plan.  
-  security_groups =  []
```
anjmao, line 68: Maybe keep example as `override_security_groups = []`

aldor007: This will remove whole SGs for the cluster.  Default is null, empty array will set SGs to empty array

PR #62, opened 2022-05-09
feat: handle cluster with empty credentials on destroy

aldor007:
Refactored code moved common parts for gke_cluster and eks_cluster to
common file, added description to eks_cluster resource so we can use doc
generator


matas-cast (approved):

matas-cast:
https://www.terraform.io/language/meta-arguments/depends_on

recommended to always add comment with `depends_on`. I think it would be nice since we had issues without it

PR #70, opened 2022-05-24
Feat/e2e poc

aldor007:


.github/workflows/e2e.yaml
```diff
@@ -0,0 +1,49 @@
...
+  workflow_run:
+    workflows: [ "Release" ]
+    types:
+      - completed
+    branches:
+      - 'master'
+  push:
+    paths-ignore:
+      - '**.md'
+      - 'examples/**'
```
aldor007, line 14: Remove me

e2e/config/config.go
```diff
@@ -0,0 +1,20 @@
+package config
+
+import (
+	"github.com/kelseyhightower/envconfig"
+)
+
+type Config struct {
+	GKEWorkspace string `envconfig:"GKE_WORKSPACE"`
+	APIURL       string `envconfig:"CASTAI_URL" default:"https://api.cast.ai"`
+	Token        string `envconfig:"CASTAI_TOKEN" defualt:""`
```
matas-cast, line 10:
```suggestion
	Token        string `envconfig:"CASTAI_TOKEN" default:""`
```

matas-cast: Is the default even needed? 🤔

e2e/.env.example
```diff
@@ -0,0 +1,3 @@
+GKE_WORKSPACE=
```
matas-cast, line 1:
```suggestion
E2E_GKE_WORKSPACE=
```
I think this should be added to the example since you're using E2E prefix?

e2e/gke_test.go
```diff
@@ -0,0 +1,70 @@
...
+
+func TestTerraformGKEOnboarding(t *testing.T) {
+	terraformOptions := terraform.WithDefaultRetryableErrors(t, &terraform.Options{
+		TerraformDir: "./tests/gke_cluster_zonal",
+	})
+
+	r := require.New(t)
+	ctx := context.Background()
+	defer func() {
+		if !t.Failed() {
```
matas-cast, line 23: why destroy only on success? 🤔

aldor007: It is easier to debug if you don't need to recreate the whole cluster (it takes 30-40 minutes for whole tests to be executed)

e2e/helpers.go
```diff
@@ -0,0 +1,57 @@
...
+	"io/ioutil"
+	"net/http"
+	"time"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/logging"
+
+	"github.com/castai/terraform-provider-castai/castai/sdk"
+)
+
+func createClient(apiURL, apiToken string) (*sdk.ClientWithResponses, error) {
```
matas-cast, line 15: this is a bit duplicate with the castai config 🤔 maybe we could re-use the client creation part?

e2e/tests/gke_cluster_zonal/gke.tf
```diff
@@ -0,0 +1,34 @@
...
+  horizontal_pod_autoscaling = true
+  filestore_csi_driver       = false
+
+  node_pools = [
+    {
+      name               = "default-node-pool"
+      machine_type       = "e2-medium"
+      node_locations     = var.cluster_zones[0]
+      min_count          = 1
+      max_count          = 100
```
matas-cast, line 23: it shouldn't autoscale but maybe let's reduce it for safety :D

PR #74, opened 2022-05-27
fix: remove cluster from state if it has archived status

aldor007:


castai/common.go
```diff
@@ -95,3 +96,26 @@ func resourceCastaiPublicCloudClusterDelete(ctx context.Context, data *schema.Re
 
 	return nil
 }
+
+func fetchClusterData(ctx context.Context, client *sdk.ClientWithResponses, data *schema.ResourceData) (*sdk.ExternalClusterAPIGetClusterResponse, error) {
+	resp, err := client.ExternalClusterAPIGetClusterWithResponse(ctx, data.Id())
+	if err != nil {
+		return nil, err
+	} else if resp.StatusCode() == http.StatusNotFound {
```
anjmao, line 104: else is redundant make makes code a bit harder to read
```
if err != nil {
		return nil, err
}

if resp.StatusCode() == http.StatusNotFound {}

File: castai/common.go [hunk 2]
```diff
@@ -95,3 +96,26 @@ func resourceCastaiPublicCloudClusterDelete(ctx context.Context, data *schema.Re
...
+		return nil, nil
+	}
+
+	if checkErr := sdk.CheckOKResponse(resp, err); checkErr != nil {
+		return nil, checkErr
+	}
+
+	if resp.JSON200 != nil && toString(resp.JSON200.Status) == sdk.ClusterStatusArchived {
+		log.Printf("[WARN] Removing cluster %s from state because it is archived in CAST AI", data.Id())
+		data.SetId("")
```
anjmao, line 116: I think fetchClusterData should not change terraform state, maybe add it to another place https://github.com/castai/terraform-provider-castai/blob/master/castai/resource_castai_eks_cluster.go#L211

aldor007: My intention was to not duplicate this code in 3 places(AKS, EKS, GKE). I would think about a better name for this method or change the approach.

varnastadeus (approved):

PR #86, opened 2022-08-29
feat: node configurations

varnastadeus:


castai/utils.go
```diff
@@ -17,6 +19,36 @@ func toStringPtr(value string) *string {
...
+		out[k] = v.(string)
+	}
+	return out
+}
+
+func toStringList(items []interface{}) []string {
+	out := make([]string, 0, len(items))
+	for _, v := range items {
+		val, ok := v.(string)
+		if ok && val != "" {
```
fals, line 45: maybe check whitespace?

fals (approved):

castai/utils.go
varnastadeus: there is validation on the field itself `validation.ToDiagFunc(validation.StringIsNotWhiteSpace)`

PR #96, opened 2022-10-06
fix: log error when updating autoscaler policy

aldor007:
image

### DIFF
--- a/castai/resource_castai_autoscaler_test.go
+++ b/castai/resource_castai_autoscaler_test.go
@@ -192,6 +192,113 @@ func TestAutoscalerResource_PoliciesUpdateAction(t *testing.T) {
 	r.True(policiesUpdated)
 }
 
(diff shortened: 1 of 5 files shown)


castai/resource_castai_autoscaler_test.go
```diff
@@ -183,6 +183,7 @@ func TestAutoscalerResource_PoliciesUpdateAction(t *testing.T) {
 
 			return &http.Response{
 				StatusCode: 200,
+				Body:       io.NopCloser(bytes.NewReader([]byte(""))),
```
fals, line 186: I would be good to have a triangulation, right now this test case only asserts with we are successful with the request to PoliciesAPIUpsertClusterPoliciesWithResponse, but we have no unit test for the error scenario

aldor007: Will add an error test case

fals (approved):

PR #105, opened 2022-11-04
feat: include version in user agent header

varnastadeus:


aldor007 (approved):

main.go
```diff
@@ -1,13 +1,21 @@
...
 
 import (
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/plugin"
 
 	"github.com/castai/terraform-provider-castai/castai"
 )
 
+var (
+	commit  = "undefined"
```
fals, line 11: is this intended, commit is not used anywhere

varnastadeus: it is set during release, kept it for visibility

PR #107, opened 2022-11-07
feat: allow to import existing node configs

varnastadeus:


aldor007 (approved):
LGTM but can you add an example/doc on how to use the import method?

PR #109, opened 2022-11-08
Remove cluster_token resource in favour of cluster attribute

andrejatcastai:


castai/resource_eks_cluster.go
```diff
@@ -57,6 +57,12 @@ func resourceEKSCluster() *schema.Resource {
 				ValidateDiagFunc: validation.ToDiagFunc(validation.StringIsNotWhiteSpace),
 				Description:      "AWS region where the cluster is placed",
 			},
+			FieldClusterToken: {
```
varnastadeus, line 60: Can we introduce this field without braking change and then do removal of `cluster_token` resource? This way users will be able to use this field without migrating to major version

PR #110, opened 2022-11-08
feat: add support for advanced node configuration options

varnastadeus:


andrejatcastai (approved):

examples/resources/node_configuration_default/resource.tf
```diff
@@ -0,0 +1,5 @@
+# Set node configuration as default one.
+resource "castai_node_configuration_default" "test" {
```
andrejatcastai, line 2: Why is it 'test'? :)

varnastadeus: it just and example

PR #126, opened 2023-01-03
fix: remove token creation from read path

varnastadeus:


castai/resource_aks_cluster.go
```diff
@@ -128,17 +129,6 @@ func resourceCastaiAKSClusterRead(ctx context.Context, data *schema.ResourceData
 			return diag.FromErr(fmt.Errorf("setting region: %w", err))
 		}
 	}
-	clusterID := *resp.JSON200.Id
-
-	if _, ok := data.GetOk(FieldClusterToken); !ok {
-		tkn, err := createClusterToken(ctx, client, clusterID)
```
aldor007, line 134: This is the only place where we create cluster token for AKS. We should move it to `resourceCastaiAKSClusterCreate`

varnastadeus: thx, didn't noticed it 🙏

aldor007 (approved):

PR #128, opened 2023-01-03
Update examples

Narunas-K:
Update examples to follow the similar structure.


examples/aks/aks_cluster/variables.tf
```diff
@@ -0,0 +1,35 @@
...
+# Variables required for connecting EKS cluster to CAST AI
+variable "castai_api_token" {
+  type = string
+  description = "CAST AI API token created in console.cast.ai API Access keys section"
+}
+
+variable "delete_nodes_on_disconnect" {
+  type        = bool
+  description = "Optional parameter, if set to true - CAST AI provisioned nodes will be deleted on cluster disconnection."
+  default     = true
```
aldor007, line 28: Not sure if this should be set to true. Is good for our testing etc but in production it should be set to `false`

examples/eks/eks_cluster_assumerole/castai.tf
```diff
@@ -1,46 +1,34 @@
...
-  cluster_id = castai_eks_clusterid.cluster_id.id
+provider "helm" {
+  kubernetes {
+    host                   = module.eks.cluster_endpoint
+    cluster_ca_certificate = base64decode(module.eks.cluster_certificate_authority_data)
+      exec {
+        api_version = "client.authentication.k8s.io/v1beta1"
+        command     = "aws"
+        # This requires the awscli to be installed locally where Terraform is executed.
+        args = ["eks", "get-token", "--cluster-name", module.eks.cluster_name]
```
aldor007, line 26: Why we can't use `data.aws_eks_cluster_auth.eks.token`?

examples/eks/eks_cluster_assumerole/variables.tf
```diff
@@ -0,0 +1,34 @@
+# EKS module variables.
+variable "cluster_name" {
+  type        = string
+  description = "EKS cluster name in AWS account."
+}
+
+variable "cluster_version" {
+  type        = string
+  description = "EKS cluster name version."
+  default     = "1.23"
```
aldor007, line 10: Maybe we can update the version of k8s?

examples/gke/gke_cluster_autoscaler/gke.tf
```diff
@@ -0,0 +1,36 @@
...
+  horizontal_pod_autoscaling = true
+  filestore_csi_driver       = false
+
+  node_pools = [
+    {
+      name               = "default-node-pool"
+      machine_type       = "e2-medium"
+      min_count          = 2 # has to be >=2 to successfully deploy CAST AI controller.
+      max_count          = 10
+      autoscaling        = true
```
aldor007, line 25: We should disable autoscaling for default GKE node pool

examples/aks/aks_cluster/variables.tf
Narunas-K: Yeah, that's true. However, I have set it to `true` for the case if the one would like to simply run Terraform apply on our examples in their cloud account and later run `terraform destroy`. The one would be sure that all infrastructure created by this example is removed. If leaving it to `false` I expect that not so advanced user could miss CAST AI created nodes and this would create bad experience for user for incurring cost which wasn't expected.

WDYT of leaving it `true` but update description that for Production use it is recommended to set it to false?

examples/eks/eks_cluster_assumerole/castai.tf
Narunas-K: `data.aws_eks_cluster_auth.eks.token` can only be used if EKS cluster exists. If `terraform apply` is run when cluster does not exist, Terraform fails with error that it cannot receive this data source as it doesn't exist. 
The same approach is used in AWS community module - https://github.com/terraform-aws-modules/terraform-aws-eks/blob/master/examples/complete/main.tf#L5

varnastadeus (commented):
regenerate docs, I believe some of the examples were referenced in tf docs previously

examples/eks/eks_cluster_assumerole/variables.tf
Narunas-K: Sure, will update it

examples/aks/aks_cluster/variables.tf
aldor007: I'm fine with adding the recommendation in description

Narunas-K: Updated as discussed.

examples/eks/eks_cluster_assumerole/variables.tf
Narunas-K: Updated.

examples/gke/gke_cluster_autoscaler/gke.tf
Narunas-K: Disabled for now.

aldor007 (approved):

PR #136, opened 2023-01-24
GKE max_pods_per_node in node configuration

fals:
- Adding max_pods_per_node in node configuration for GKE provider


aldor007 (approved):
LGTM

PR #144, opened 2023-02-08
Omit max zero values in node templates

zilvinasu:


zilvinasu:
https://github.com/hashicorp/terraform-plugin-sdk/issues/90

aldor007 (approved):

PR #147, opened 2023-02-22
Introduce Volume type and Volume IOPS parameters to EKS node configuration

Narunas-K:
…ation

### DIFF
--- a/castai/resource_node_configuration.go
+++ b/castai/resource_node_configuration.go
@@ -161,8 +161,20 @@ func resourceNodeConfiguration() *schema.Resource {
 						"key_pair_id": {
 							Type:             schema.TypeString,
 							Optional:         true,
-							Description:      "AWS key pair ID to be used for provisioned nodes. Has priority over ssh_public_key",
+							Description:      "AWS key pair ID to be used for CAST provisioned nodes. Has priority over ssh_public_key",
 							ValidateDiagFunc: castval.ValidKeyPairFormat(),
+						},
+						"volume_type": {
+							Type:             schema.TypeString,
+							Optional:         true,
+							Description:      "AWS EBS volume type to be used for CAST provisioned nodes",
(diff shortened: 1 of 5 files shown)


castai/resource_node_configuration.go
```diff
@@ -161,9 +161,21 @@ func resourceNodeConfiguration() *schema.Resource {
 						"key_pair_id": {
 							Type:             schema.TypeString,
 							Optional:         true,
-							Description:      "AWS key pair ID to be used for provisioned nodes. Has priority over ssh_public_key",
+							Description:      "AWS key pair ID to be used for CAST provisioned nodes. Has priority over ssh_public_key",
 							ValidateDiagFunc: castval.ValidKeyPairFormat(),
 						},
+						"volume_type": {
```
fals, line 167: Would be nice to also add include in the unit test those new attributes -> TestAccResourceNodeConfiguration_basic

castai/resource_node_configuration.go
```diff
@@ -161,9 +161,21 @@ func resourceNodeConfiguration() *schema.Resource {
...
+							Type:             schema.TypeString,
+							Optional:         true,
+							Description:      "AWS EBS volume type to be used for CAST provisioned nodes",
+							ValidateDiagFunc: castval.ValidEc2VolumeType(),
```
aldor007, line 171:
```suggestion
							ValidateDiagFunc: validation.ToDiagFunc(validation.StringInSlice([]string{"gp3", "io1", "io2"}, true)),
							
```

WDYT?

Narunas-K: Makes sense! That was I was looking for.

castai/resource_node_configuration.go
Narunas-K: Thanks, good point. Included as suggested.

aldor007 (approved):

PR #165, opened 2023-05-12
Add azure CNI example & clean

mjudeikis:
Add azure CNI example and clean tf.vars for local dev


examples/aks/aks_cluster_azure_cni/variables.tf
```diff
@@ -0,0 +1,52 @@
...
+variable "cluster_network_provider" {
+  type        = string
+  description = "AKS cluster network provider"
+  default     = "azure"
+}
+
+variable "cluster_network_provider_mode" {
+  type        = string
+  description = "AKS cluster network provider mode"
+  default     = ""
```
aldor007, line 27: Shouldn't it be `Overlay` as a default given the example for azure cni?

examples/aks/aks_cluster_azure_cni/variables.tf
```diff
@@ -0,0 +1,52 @@
...
+  default     = "1.23"
+}
+
+variable "cluster_network_provider" {
+  type        = string
+  description = "AKS cluster network provider"
+  default     = "azure"
+}
+
+variable "cluster_network_provider_mode" {
```
fals, line 24: Would be good having a   validation for this one, because they are predefined values in the inner package

examples/aks/aks_cluster_azure_cni/variables.tf
```diff
@@ -0,0 +1,52 @@
...
+variable "cluster_version" {
+  type        = string
+  description = "AKS cluster version."
+  default     = "1.23"
+}
+
+variable "cluster_network_provider" {
+  type        = string
+  description = "AKS cluster network provider"
+  default     = "azure"
```
fals, line 21: Maybe also for this one a validation?

examples/aks/aks_cluster_azure_cni/variables.tf
mjudeikis: CNI has 2 modes. "" and Overlay. So by default its "", but you can do "Overlay" if you chose to. Considering its new feature I wanted to keep it not default now.

examples/aks/aks_cluster_azure_cni/variables.tf
mjudeikis: Will add this with next batch :)

PR #176, opened 2023-06-15
Support EKS imds_hop_limit in node config

fals:
- imds_hop_limit added to EKS node configuration fild


castai/resource_node_configuration.go
```diff
@@ -197,6 +197,11 @@ func resourceNodeConfiguration() *schema.Resource {
 							Optional:    true,
 							Description: "Allow IMDSv1, the default is true",
 						},
+						"imds_hop_limit": {
+							Type:        schema.TypeInt,
+							Optional:    true,
+							Description: "Allow configure the IMDSv2 hop limit, the default is 2",
```
aldor007, line 203: Maybe let's set default also in terraform? Like for example cpuDiskratio?

aldor007 (approved):

castai/resource_node_configuration.go
fals: done

PR #177, opened 2023-06-19
CO-2326: Add the execution conditions to the scheduled rebalancing

kiriloman:
The execution conditions define conditions that must be met during the execution of the plan in order for it to run until the end. 

For example, if the execution conditions are enabled but the percentage of achieved savings is not met when the new nodes are created, the rebalancing plan will not proceed further and will rollback the created nodes. This is specially useful when, initially, the rebalancing plan predicts X% of savings, but due to certain constraints, that are known during the runtime, like unavailability of instance types, it creates instance types that are pricier than predicted.

In this MR, the execution conditions allow to control the minimum % of the predicted savings that must be achieved in order to continue with the rebalancing plan.

### DIFF
--- a/castai/resource_rebalancing_schedule.go
+++ b/castai/resource_rebalancing_schedule.go
@@ -95,6 +95,28 @@ func resourceRebalancingSchedule() *schema.Resource {
 							ValidateDiagFunc: validation.ToDiagFunc(validation.IntAtLeast(1)),
 							Description:      "Minimum number of nodes that should be kept in the cluster after rebalancing.",
 						},
+						"execution_conditions": {
+							Type:     schema.TypeList,
+							MaxItems: 1,
+							Optional: true,
+							Elem: &schema.Resource{
+								Schema: map[string]*schema.Schema{
+									"enabled": {
+										Type:        schema.TypeBool,
+										Required:    true,
+										Description: "Enables or disables the execution conditions.",
+									},
+									"achieved_savings_percentage": {
+										Type:             schema.TypeInt,
+										Optional:         true,
+										Default:          0,
+										ValidateDiagFunc: validation.ToDiagFunc(validation.IntBetween(0, 100)),
+										Description: "The percentage of the predicted savings that must be achieved in order to fully execute the plan." +
+											"If the savings are not achieved after creating the new nodes, the plan will fail and delete the created nodes.",
+									},
+								},
+							},
+						},
 						"selector": {
 							Type:             schema.TypeString,
 							Optional:         true,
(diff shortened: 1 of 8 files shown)


castai/resource_rebalancing_schedule.go
```diff
@@ -220,11 +242,22 @@ func stateToSchedule(d *schema.ResourceData) (*sdk.ScheduledrebalancingV1Rebalan
 		if err != nil {
 			return nil, fmt.Errorf("parsing selector: %w", err)
 		}
+
+		var executionConditions *sdk.ScheduledrebalancingV1ExecutionConditions
```
viktorasm, line 246: skip some boilerplate here with `toSection`, which does the cast & selection of the first element

castai/resource_rebalancing_schedule.go
```diff
@@ -250,8 +283,18 @@ func scheduleToState(schedule *sdk.ScheduledrebalancingV1RebalancingSchedule, d
 		"node_ttl_seconds":   schedule.LaunchConfiguration.NodeTtlSeconds,
 		"num_targeted_nodes": schedule.LaunchConfiguration.NumTargetedNodes,
 	}
+
 	if schedule.LaunchConfiguration.RebalancingOptions != nil {
 		launchConfig["rebalancing_min_nodes"] = schedule.LaunchConfiguration.RebalancingOptions.MinNodes
+
+		if schedule.LaunchConfiguration.RebalancingOptions.ExecutionConditions != nil {
```
viktorasm, line 290: introduce a temporary variable for `schedule.LaunchConfiguration.RebalancingOptions.ExecutionConditions` to reduce copy/paste in the mapping

viktorasm (approved):

castai/resource_rebalancing_schedule.go
```diff
@@ -95,6 +95,28 @@ func resourceRebalancingSchedule() *schema.Resource {
...
+									"enabled": {
+										Type:        schema.TypeBool,
+										Required:    true,
+										Description: "Enables or disables the execution conditions.",
+									},
+									"achieved_savings_percentage": {
+										Type:             schema.TypeInt,
+										Optional:         true,
+										Default:          0,
+										ValidateDiagFunc: validation.ToDiagFunc(validation.IntAtLeast(0)),
```
zilvinasu, line 113: should validate max as well ?

zilvinasu (changes requested):

castai/resource_rebalancing_schedule.go
kiriloman: Wanted to use that one, but I don't think it supports returning nested blocks. So doing something like:

```
toSection(d, "launch_configuration.execution_conditions")
```

won't return the `execution_conditions`

PR #180, opened 2023-06-29
Add support for GKE boot disk type in node configuration

apasyniuk:


fals (approved):

Narunas-K (changes requested):
As new field was added, let's include it to acceptance test as well. E.g. https://github.com/castai/terraform-provider-castai/blob/master/castai/resource_node_configuration_test.go#L44-L48

Narunas-K (approved):

PR #181, opened 2023-07-03
EKS GitOps flow example configuration

Narunas-K:


examples/eks/eks_cluster_gitops/README.md
```diff
@@ -0,0 +1,26 @@
+## EKS and CAST AI example for GitOps onboarding flow
+
+Following example shows how to onboard EKS cluster to CAST AI using GitOps flow.
```
fals, line 3: Could you add/make clear for the reader that this example is a way to manage Node Configuration, Node Templates, etc using terraform, but all castware component need to be installed manually by other means.

fals (approved):

examples/eks/eks_cluster_gitops/variables.tf
```diff
@@ -0,0 +1,60 @@
...
+variable "aws_assume_role_arn" {
+  type        = string
+  description = "Arn of the role to be used by CAST AI for IAM access"
+  default     = null
+}
+
+variable "subnets" {
+  type = list(string)
+  description = "Subnet IDs used by CAST AI to provision nodes"
+  #  default = ["subnet-12345678abcdefghi", "subnet-12345678abcdefghz"]
```
aldor007, line 32: Do we need those commented values?

examples/eks/eks_cluster_gitops/README.md
```diff
@@ -0,0 +1,26 @@
...
+are installed by the user outside of Terraform using other means e.g. ArgoCD.
+
+Steps to take to successfully onboard EKS cluster to CAST AI using GitOps flow:
+Prerequisites:
+- CAST AI account
+- Obtained CAST AI [API Access key](https://docs.cast.ai/docs/authentication#obtaining-api-access-key) with Full Access
+1. Configure `tf.vars.example` file with required values. If EKS cluster is already managed by Terraform you could instead directly reference those resources.
+2. Run `terraform init`
+3. Run `terraform apply` and make a note of `cluster_id` and `cluster_token` output values. At this stage you would see that your cluster is in `Connecting` state in CAST AI console
+4. Install CAST AI components using Helm. Use `cluster_id` and `cluster_token` values to configure Helm releases:
```
aldor007, line 14: Can we add here an example of how to use argo/flux/helm (one of them)

aldor007:
Also, I think we should create `README.md` in `examples/eks/README.md` as we have there a few different scenarios. It would be easier to navigate

examples/eks/eks_cluster_gitops/README.md
Narunas-K: Thanks! Added

Narunas-K:
> Also, I think we should create `README.md` in `examples/eks/README.md` as we have there a few different scenarios. It would be easier to navigate

Good idea, added it as suggested.

PR #183, opened 2023-07-06
feat: handle phase1 mode

aldor007:


fals:
We have a README.md in every single example explaining what this terraform is for. Could you please also add for this one?

PR #185, opened 2023-07-07
Introduce optional read-only mode examples

Narunas-K:


examples/eks/eks_cluster_optional_readonly/variables.tf
```diff
@@ -0,0 +1,45 @@
...
+}
+
+variable "aws_assume_role_arn" {
+  type        = string
+  description = "Arn of the role to be used by CAST AI for IAM access"
+  default     = null
+}
+
+variable "readonly" {
+  type = bool
```
aldor007, line 43: I would add some description for this variable

aldor007 (approved):

examples/eks/eks_cluster_optional_readonly/variables.tf
Narunas-K: Updated

PR #187, opened 2023-07-11
Change imds_v1 node_configuration field for EKS default value to true.

apasyniuk:
Change imds_v1 node_configuration field for EKS default value to true.
Remove deprecated fields: castai_eks_clusterid, castai_cluster_token


castai/resource_node_configuration.go
```diff
@@ -193,8 +193,9 @@ func resourceNodeConfiguration() *schema.Resource {
 							ValidateDiagFunc: validation.ToDiagFunc(validation.IntBetween(125, 1000)),
 						},
 						"imds_v1": {
-							Type:        schema.TypeBool,
-							Optional:    true,
+							Type:     schema.TypeBool,
+							Optional: true,
+							Default:  true,
```
fals, line 198: 1. We have to gen-docs again.

2. Could you improve the description of this field?

`When the value is true both IMDSv1 and IMDSv2 are enabled. Setting the value to false disables permanently  IMDSv1 and might affect legacy workloads running on the node created with this configuration. The default is true if the flag isn't provided`

fals (commented):
1. We have to gen-docs again.

2. Could you improve the description of this field?

`When the value is true both IMDSv1 and IMDSv2 are enabled. Setting the value to false disables permanently  IMDSv1 and might affect legacy workloads running on the node created with this configuration. The default is true if the flag isn't provided`

castai/resource_node_configuration.go
apasyniuk: Ad1 I regenerated docs after my change but it changed nothing.
Ad2 Updated description and regenerated docs.

fals (approved):
LGTM

varnastadeus (approved):

PR #188, opened 2023-07-13
GKE GitOps flow example configuration

davex98:


examples/gke/gke_cluster_gitops/main.tf
```diff
@@ -0,0 +1,32 @@
+resource "castai_gke_cluster" "my_castai_cluster" {
```
aldor007, line 1:
```suggestion
resource "castai_gke_cluster" "this" {
```

We will have only one resource with that name :)

aldor007 (approved):

examples/gke/gke_cluster_gitops/main.tf
davex98: Valid point, thanks! 🚀

PR #194, opened 2023-07-21
feat: rename rebalancing option keep_drain_timeout_nodes => evict_gracefully

viktorasm:


castai/resource_node_template.go
```diff
@@ -605,7 +607,9 @@ func getNodeTemplateByName(ctx context.Context, data *schema.ResourceData, meta
 	nodeTemplateName := data.Id()
 
 	log.Printf("[INFO] Getting current node templates")
-	resp, err := client.NodeTemplatesAPIListNodeTemplatesWithResponse(ctx, clusterID)
+	resp, err := client.NodeTemplatesAPIListNodeTemplatesWithResponse(ctx, clusterID, &sdk.NodeTemplatesAPIListNodeTemplatesParams{
+		IncludeDefault: lo.ToPtr(false),
```
kiriloman, line 611: This PR seems to have unrelated Node Template changes

viktorasm: API changes forced this one, can't do anything about it - could maybe separate out as another PR, but seems like a small change to pack together.

castai/resource_rebalancing_schedule.go
```diff
@@ -95,7 +96,7 @@ func resourceRebalancingSchedule() *schema.Resource {
 							ValidateDiagFunc: validation.ToDiagFunc(validation.IntAtLeast(1)),
 							Description:      "Minimum number of nodes that should be kept in the cluster after rebalancing.",
 						},
-						"keep_drain_timeout_nodes": {
```
aldor007, line 98: This is a breaking change. We shouldn't just remove the filed — first deprecation and than remove.

castai/resource_rebalancing_schedule.go
```diff
@@ -248,7 +255,10 @@ func stateToSchedule(d *schema.ResourceData) (*sdk.ScheduledrebalancingV1Rebalan
 			return nil, fmt.Errorf("parsing selector: %w", err)
 		}
 
-		keepDrainTimeoutNodes := readOptionalValue[bool](launchConfigurationData, "keep_drain_timeout_nodes")
+		evictGracefully := readOptionalValue[bool](launchConfigurationData, "evict_gracefully")
+		if evictGracefully == nil || *evictGracefully == false {
```
viktorasm, line 259: this place is a hack as TF initializes both "evict_gracefully" and "keep_drain_timeout_nodes" as part of the nested struct.

castai/resource_rebalancing_schedule.go
```diff
@@ -294,7 +304,12 @@ func scheduleToState(schedule *sdk.ScheduledrebalancingV1RebalancingSchedule, d
 
 	if schedule.LaunchConfiguration.RebalancingOptions != nil {
 		launchConfig["rebalancing_min_nodes"] = schedule.LaunchConfiguration.RebalancingOptions.MinNodes
-		launchConfig["keep_drain_timeout_nodes"] = schedule.LaunchConfiguration.RebalancingOptions.KeepDrainTimeoutNodes
+
+		if _, ok := d.GetOk("launch_configuration.keep_drain_timeout_nodes"); ok {
```
viktorasm, line 308: not even sure how to implement this part: if user uses the old field, the fill the old field, otherwise use the new field.

PR #214, opened 2023-08-10
test: Use random labels in node template acceptance tests

jansyk13:


castai/resource_node_template_test.go
```diff
@@ -371,13 +371,13 @@ func TestAccResourceNodeTemplate_basic(t *testing.T) {
 					resource.TestCheckResourceAttr(resourceName, "custom_instances_enabled", "false"),
 					resource.TestCheckResourceAttr(resourceName, "custom_label.#", "0"),
 					resource.TestCheckResourceAttr(resourceName, "custom_labels.%", "2"),
-					resource.TestCheckResourceAttr(resourceName, "custom_labels.custom-key-1", "custom-value-1"),
-					resource.TestCheckResourceAttr(resourceName, "custom_labels.custom-key-2", "custom-value-2"),
+					resource.TestCheckResourceAttr(resourceName, fmt.Sprintf("custom_labels.%s-1", rName), fmt.Sprintf("%s-1", rName)),
```
viktorasm, line 374: feels like there's some some potential for copypaste reduction

castai/resource_node_template_test.go
```diff
@@ -414,11 +414,11 @@ func TestAccResourceNodeTemplate_basic(t *testing.T) {
...
+					resource.TestCheckResourceAttr(resourceName, fmt.Sprintf("custom_labels.%s-2", rName), fmt.Sprintf("%s-2", rName)),
 					resource.TestCheckResourceAttr(resourceName, "custom_taints.#", "1"),
-					resource.TestCheckResourceAttr(resourceName, "custom_taints.0.key", "custom-taint-key-1"),
-					resource.TestCheckResourceAttr(resourceName, "custom_taints.0.value", "custom-taint-value-1"),
+					resource.TestCheckResourceAttr(resourceName, "custom_taints.0.key", fmt.Sprintf("%s-1", rName)),
+					resource.TestCheckResourceAttr(resourceName, "custom_taints.0.value", fmt.Sprintf("%s-1", rName)),
```
viktorasm, line 421: taint key and value was different values before

castai/resource_node_template_test.go
```diff
@@ -414,11 +414,11 @@ func TestAccResourceNodeTemplate_basic(t *testing.T) {
 					resource.TestCheckResourceAttr(resourceName, "custom_instances_enabled", "false"),
 					resource.TestCheckResourceAttr(resourceName, "custom_label.#", "0"),
 					resource.TestCheckResourceAttr(resourceName, "custom_labels.%", "2"),
-					resource.TestCheckResourceAttr(resourceName, "custom_labels.custom-key-1", "custom-value-1"),
-					resource.TestCheckResourceAttr(resourceName, "custom_labels.custom-key-2", "custom-value-2"),
+					resource.TestCheckResourceAttr(resourceName, fmt.Sprintf("custom_labels.%s-1", rName), fmt.Sprintf("%s-1", rName)),
```
viktorasm, line 417: by using "resourceName-1" for everything you're reducing test strength because now implementation can be seriously messed up and test might pass because every field expects same value.

jansyk13: You are right, I thought about it as well. But my thinking was that if labels/taints get mixed up. It should be caught by different test.

I have added fixup with different values for each attribute.

aldor007 (approved):

castai/resource_node_template_test.go
jansyk13: Fixed

PR #226, opened 2023-09-18
fix: remove unsupported PreferNoSchedule taint effect option

laimonasr:


castai/resource_node_template.go
```diff
@@ -353,7 +352,7 @@ func resourceNodeTemplate() *schema.Resource {
 							Type:     schema.TypeString,
 							Default:  TaintEffectNoSchedule,
 							ValidateDiagFunc: validation.ToDiagFunc(
-								validation.StringInSlice([]string{TaintEffectNoSchedule, TaintEffectNoExecute, TaintEffectPreferNoSchedule}, false),
+								validation.StringInSlice([]string{TaintEffectNoSchedule, TaintEffectNoExecute}, false),
```
fals, line 355: should generate the docs again to update the fields

castai/resource_node_template.go
```diff
@@ -909,7 +908,7 @@ func flattenCustomTaints(taints *[]sdk.NodetemplatesV1Taint) []map[string]string
 			t[FieldValue] = toString(v)
 		}
 		if e := taint.Effect; e != nil {
-			t[FieldEffect] = toString(e)
+			t[FieldEffect] = string(*e)
```
fals, line 911: why changing this? The toString is guarantee no nil pointer

castai/resource_node_template.go
laimonasr: Documentation is regenerated but there are no changes probably because validation information is not used when generating it

castai/resource_node_template.go
laimonasr: `toString` accepts `*string` but the type of `taint.Effect` is `*NodetemplatesV1TaintEffect` which even though is alias cannot be passed as `*string` without casting. There's a check `e != nil` so it's not going to be nil. I can make it `toString((*string)(e))` if you think it it's better?

castai/resource_node_template.go
laimonasr: Will add allowed values to the field description as we do for some other fields

aldor007 (approved):

PR #228, opened 2023-10-06
Refactor SDK 

mjudeikis:
* Refactors SDK to use new generators
* Refactors UserAPI sdk
* Moves code to use SDK types types `WithResponse` over parsing body (not everywhere but in most cases).


mjudeikis:
Generate SDK fails until public swagger changes too.

PR #236, opened 2023-10-24
Handle empty instance_families

linkas45:


castai/resource_node_template.go
```diff
@@ -951,10 +951,10 @@ func toTemplateConstraints(obj map[string]any) *sdk.NodetemplatesV1TemplateConst
 	if v, ok := obj[FieldNodeTemplateFallbackRestoreRateSeconds].(int); ok {
 		out.FallbackRestoreRateSeconds = toPtr(int32(v))
 	}
-	if v, ok := obj[FieldNodeTemplateGpu].([]any); ok && len(v) > 0 {
+	if v, ok := obj[FieldNodeTemplateGpu].([]any); ok && len(v) > 0 && v[0] != nil {
```
laimonasr, line 954: Instead of nil checks I would suggest to use type assertion because it's possible that the value in the slice is still not of type `map[string]any`. Something like this:
```
val, ok := v[0].(map[string]any)
if ok {
    out.Gpu = toTemplateConstraintsGpuConstraints(val)
}
```

linkas45: thx, updated

laimonasr (approved):

Makefile
```diff
@@ -5,8 +5,8 @@ default: build
 init-examples:
 	@echo "==> Creating symlinks for example/ projects to terraform-provider-castai binary"; \
 	TF_PROVIDER_FILENAME=terraform-provider-castai; \
-	GOOS=`go tool dist env | awk -F'=' '/^GOOS/ { print $$2}' | tr -d '"'`; \
-	GOARCH=`go tool dist env | awk -F'=' '/^GOARCH/ { print $$2}' | tr -d '"'`; \
+	GOOS=`go tool dist env | awk -F'=' '/^GOOS/ { print $$2}' | tr -d '";'`; \
```
aldor007, line 8: Why do we need this change?

linkas45: in my local environment `go tool dist env` provides list with quotes and semicolons. Like this:
`
GOHOSTARCH="arm64";
GOHOSTOS="darwin";
`

linkas45: So this change remove semicolons from the name too

aldor007 (approved):

PR #242, opened 2023-11-15
chore: regenerate SDK

davex98:


castai/resource_organization_members.go
```diff
@@ -90,11 +90,11 @@ func resourceOrganizationMembersCreate(ctx context.Context, data *schema.Resourc
 			// Person that creates a new organization is automatically the owner
 			// of it. That's why when creating this resource we would like to skip
 			// re-creating that user because it would fail.
-			if currentUserResp.JSON200.Email == email {
+			if *currentUserResp.JSON200.Email == email {
```
fals, line 93: should we trust this pointer? maybe use lo.FromPtr to sleep at night?

davex98: When we return the current user we always return the email but I have added  `lo.FromPtr` just in case

fals (approved):
LGTM

PR #253, opened 2023-12-07
feat: Add pod-pinner to CAST AI onboarding

andrejatcastai:


examples/aks/aks_cluster/variables.tf
```diff
@@ -27,6 +27,11 @@ variable "castai_api_token" {
   description = "CAST AI API token created in console.cast.ai API Access keys section"
 }
 
+variable "castai_grpc_url" {
+  type        = string
+  description = "CAST AI gRPC URL"
+}
+
```
julgircast, line 34: Probably need to provide default URL if this is a mandatory variable.

andrejatcastai: Very good catch. Added

aldor007 (approved):

PR #265, opened 2024-01-03
fix: detect external autoscaler policy changes

laimonasr:


castai/resource_autoscaler.go
```diff
@@ -5,12 +5,12 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	jsonpatch "github.com/evanphx/json-patch"
 	"io"
 	"log"
 	"net/http"
 	"time"
 
-	jsonpatch "github.com/evanphx/json-patch"
```
viktorasm, line 13: nitpick: this moved to the wrong group for no reason

castai/utils.go
```diff
@@ -107,3 +111,12 @@ func toNilList[T any](l *[]T) *[]T {
 	}
 	return l
 }
+
+func normalizeJSON(bytes []byte) ([]byte, error) {
```
viktorasm, line 115: what is this supposed to achieve?

castai/resource_autoscaler.go
```diff
@@ -74,6 +75,23 @@ func resourceCastaiAutoscalerDelete(ctx context.Context, data *schema.ResourceDa
...
+	clusterId := getClusterId(d)
+	if clusterId == "" {
+		return nil
+	}
+
+	policies, err := getChangedPolicies(ctx, d, meta, clusterId)
+	if err != nil {
+		return err
+	}
+	if policies == nil {
```
viktorasm, line 88: probably a redundant check as method does not return nil,nil

viktorasm: ah, nevermind, I can see one path has a nil,nil result.

viktorasm (approved):

castai/utils.go
laimonasr: Umarshal and marshal produces the same json string when the input json has differently ordered keys. Terraform would detect whitespace changes when json values are the same but their keys are ordered differently so this function normalizes the key order

castai/resource_autoscaler.go
laimonasr: Moved it back

castai/utils.go
viktorasm: can you stress test it with a sample JSON? I think it uses maps as an intermediate holding structure, and those do not always result in same iteration order, e.g. tests based on map key order tend to be flaky for this reason.

laimonasr: Added a unit test with a few test cases

aldor007 (approved):

PR #269, opened 2024-01-15
Added aws profile variable when trying to install cast.ai using the eks_cluster_existing example

Aym3nTN:
This is to give the end user the possibility to use a different AWS Profile value other than 'default' and avoiding the use of 
`export AWS_DEFAULT_PROFILE=my-profile-xyz`
Otherwise, one might get the following error: `Error: Kubernetes cluster unreachable: the server has asked for the client to provide credentials`


aldor007:
@Aym3nTN thank you for the contribution!

PR #281, opened 2024-02-28
NOJIRA: Add is_enabled value to node templates in examples

linkas45:


castai/sdk/api.gen.go
```diff
@@ -51,6 +51,13 @@ const (
 	Ssd     CastaiInventoryV1beta1StorageInfoDeviceType = "ssd"
 )
 
+// Defines values for CastaiSsoV1beta1OIDCType.
+const (
+	TYPEBACKCHANNEL  CastaiSsoV1beta1OIDCType = "TYPE_BACK_CHANNEL"
+	TYPEFRONTCHANNEL CastaiSsoV1beta1OIDCType = "TYPE_FRONT_CHANNEL"
+	TYPEUNKNOWN      CastaiSsoV1beta1OIDCType = "TYPE_UNKNOWN"
+)
+
```
aweris, line 60: [question] why do we have generated file changes in this PR?

aweris (approved):
Other than gen files LGTM. Separating it from this PR would be great, if it's possible, and not related to the example changes.

castai/sdk/api.gen.go
linkas45: without this pipeline fails with error "SDK is not up to date, regenerate it by running 'make generate-sdk' and commit your changes."

aweris: yes, weird. I wasn't able to find any broken merge PR for these files. it could be some dependency change or some 3rd party change triggered this generation.

PR #288, opened 2024-03-26
docs: Remove Dedicated Node Affinity FF

jansyk13:


castai/sdk/api.gen.go
```diff
@@ -1825,7 +1825,7 @@ type NodeconfigV1TargetGroup struct {
 	Arn *string `json:"arn,omitempty"`
 
 	// Port of the target group.
-	Port *int32 `json:"port,omitempty"`
+	Port *int32 `json:"port"`
```
aweris, line 1828: is this intentional? it felt like it should be here

aweris (approved):
LGTM, please remove `api.gen.go` if possible to separate PR

PR #291, opened 2024-04-10
feat: add support for setting node affinity selectors for dedicated n…

mikenorgate:
…odes

### DIFF
--- a/castai/resource_node_template.go
+++ b/castai/resource_node_template.go
@@ -382,7 +382,7 @@ func resourceNodeTemplate() *schema.Resource {
 										Description: "Availability zone name.",
 									},
 									FieldNodeTemplateName: {
-										Optional:    true,
+										Required:    true,
 										Type:        schema.TypeString,
 										Description: "Name of node group.",
 									},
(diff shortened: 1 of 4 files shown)


castai/resource_node_template.go
```diff
@@ -368,10 +382,37 @@ func resourceNodeTemplate() *schema.Resource {
 										Description: "Availability zone name.",
 									},
 									FieldNodeTemplateName: {
-										Required:    true,
+										Optional:    true,
```
jansyk13, line 385: Intended change?

castai/resource_node_template.go
```diff
@@ -368,10 +382,37 @@ func resourceNodeTemplate() *schema.Resource {
...
 										Type:        schema.TypeString,
 										Description: "Name of node group.",
 									},
+									FieldNodeTemplateAffinityName: {
```
jansyk13, line 389: Add affinity to unit test

castai/resource_node_template.go
mikenorgate: Yes, you don't need to provide the dedicated node group name if you are using the selectors

jansyk13 (approved):

PR #296, opened 2024-04-18
fix: handle deleted scheduled rebalancing job and recreate it

jansyk13:


aweris (approved):

aweris:
question[non-blocking]: The generated file seems not related to PR.  Why do we have that change in this PR?

jansyk13:
> question[non-blocking]: The generated file seems not related to PR. Why do we have that change in this PR?

PR checks requires up to date SDK

PR #300, opened 2024-04-23
feat: Commitments and commitment assignments support

maxtwardowski:
The purpose of this PR is to ship a new resource type responsible for CAST AI's commitments and commitment assignments provisioning.


mindaugasCast:
LGTM

castai/utils.go
```diff
@@ -120,3 +125,20 @@ func normalizeJSON(bytes []byte) ([]byte, error) {
 	}
 	return json.Marshal(output)
 }
+
+func getDefaultOrganizationId(ctx context.Context, meta any) (string, error) {
```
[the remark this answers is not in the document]
maxtwardowski: Unfortunately yes. In the current `castai_reservations` resource organization ID is passed as an optional argument because the endpoints consume it (it uses InventoryAPI):
image

In the CommitmentsAPI though, we no longer have the argument meaning that the default organization ID associated with the auth token is used. So if a user has access to multiple organizations, the flow is to use multiple `castai` providers with different auth tokens.

furkhat (changes requested):
I see that the api responses converted to tf resources in an unconventional way. Other resources follow a flow of api call -> flattenXYZ/expandXYZ -> set to state. Though that mapping logic is well polished and not bad I think it would make sense to follow common convention with flatten/expand functions.
What do you think?

maxtwardowski:
As discussed during the meeting - I don't use the flatten/expand convention as I figured out that Terraform SDK uses https://github.com/mitchellh/mapstructure under the hood. So instead of mapping API types to `map[string]any` where keys are defined by constant strings, we can just use structs with `mapstructure` struct tags. In this way the decoding part is automatically done by the library and there's no need to do the mapping manually.

castai/resource_commitments_mapping.go
```diff
@@ -0,0 +1,416 @@
...
+	// commitmentResource is an interface for common management of GCP (gcpCUDResource) and Azure (azureReservationResource) resources
+	commitmentResource interface {
+		// GetCommitmentID returns the ID of the commitment in CAST AI
+		GetCommitmentID() string
+		// GetIDInCloud returns the ID of the resource in the cloud provider
+		GetIDInCloud() string
+	}
+)
+
+func (r *gcpCUDConfigResource) GetMatcher() *gcpCUDConfigMatcherResource {
```
furkhat, line 64: Could these functions stay private as well? Unless it is to satisfy some Go interface, which I think is only the case for two funcs of `commitmentResource`, right?.

castai/resource_commitments.go
```diff
@@ -0,0 +1,541 @@
...
+)
+
+func resourceCommitments() *schema.Resource {
+	return &schema.Resource{
+		ReadContext:   resourceCastaiCommitmentsRead,
+		CreateContext: resourceCastaiCommitmentsCreate,
+		UpdateContext: resourceCastaiCommitmentsUpdate,
+		DeleteContext: resourceCastaiCommitmentsDelete,
+		Importer: &schema.ResourceImporter{
+			StateContext: commitmentsStateImporter,
```
furkhat, line 36: There is a mechanism in sdk for simple cases when import can reuse read function.
```suggestion
			StateContext: schema.ImportStatePassthroughContext,
```

furkhat (approved):
Thank you for adding extensive tests. Feel free to merge after you address comments however way you think best.

castai/resource_commitments.go
maxtwardowski: Good point, I didn't know about this one

PR #302, opened 2024-04-25
Feature: Commitments support for Azure's reservations

maxtwardowski:


castai/commitments/mapping.go
```diff
@@ -201,45 +251,86 @@ func MapCUDImportToResource(
...
+
+	switch res.Plan { // normalize the values just like CAST AI's API does
+	case "P1Y":
+		res.Plan = "ONE_YEAR"
+	case "P3Y":
+		res.Plan = "THREE_YEAR"
+	case "ONE_YEAR":
+	case "THREE_YEAR":
+	default:
+		res.Plan = "ONE_YEAR"
```
ivanstankovcast, line 280: Do we need to have a default value? or return validation error?

maxtwardowski: This is what the API does so my intention here was to mimic its behaviour: https://gitlab.com/castai/kubecast/blob/15a6e46e7e23680a1104de2f55a4812f3cb3911a/services/inventory/internal/server/commitments/server.go#L535

Now when I think of it, it might be good to return an error actually.

ivanstankovcast: please, return error

ivanstankovcast (approved):

PR #306, opened 2024-05-02
refactor!: deprecate `compute_optimized` and `storage_optimized` contraints

aweris:
Deprecate `compute_optimized` and `storage_optimized` constraints and introduce `compute_optimized_status` and `storage_optimized_status` to allow empty values

### DIFF
--- a/castai/resource_node_template.go
+++ b/castai/resource_node_template.go
@@ -8,6 +8,7 @@ import (
 	"time"
 
 	"github.com/google/uuid"
+	"github.com/hashicorp/go-cty/cty"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/retry"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
@@ -20,6 +21,7 @@ import (
 const (
 	FieldNodeTemplateArchitectures                            = "architectures"
 	FieldNodeTemplateComputeOptimized                         = "compute_optimized"
+	FieldNodeTemplateComputeOptimizedStatus                   = "compute_optimized_status"
 	FieldNodeTemplateConfigurationId                          = "configuration_id"
 	FieldNodeTemplateConstraints                              = "constraints"
 	FieldNodeTemplateCustomInstancesEnabled                   = "custom_instances_enabled"
@@ -54,6 +56,7 @@ const (
 	FieldNodeTemplateSpotInterruptionPredictionsEnabled       = "spot_interruption_predictions_enabled"
 	FieldNodeTemplateSpotInterruptionPredictionsType          = "spot_interruption_predictions_type"
 	FieldNodeTemplateStorageOptimized                         = "storage_optimized"
+	FieldNodeTemplateStorageOptimizedStatus                   = "storage_optimized_status"
 	FieldNodeTemplateUseSpotFallbacks                         = "use_spot_fallbacks"
 	FieldNodeTemplateCustomPriority                           = "custom_priority"
 	FieldNodeTemplateDedicatedNodeAffinity                    = "dedicated_node_affinity"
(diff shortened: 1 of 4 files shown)


castai/resource_node_template.go
```diff
@@ -219,10 +221,20 @@ func resourceNodeTemplate() *schema.Resource {
...
 						},
 						FieldNodeTemplateStorageOptimized: {
-							Type:        schema.TypeBool,
-							Optional:    true,
-							Default:     false,
-							Description: "Storage optimized instance constraint - will only pick storage optimized nodes if true",
+							Type:          schema.TypeBool,
+							Optional:      true,
+							Default:       false,
+							Deprecated:    "Use `storage_optimized_enum` instead. Attribute will be removed in the next major release.",
```
linkas45, line 227: but we should remove it in this release as discussed, right? So if customer upgrades and have set this flag, we return an error that they have to remove this value and set enum value instead

castai/resource_node_template.go
```diff
@@ -54,6 +55,7 @@ const (
 	FieldNodeTemplateSpotInterruptionPredictionsEnabled       = "spot_interruption_predictions_enabled"
 	FieldNodeTemplateSpotInterruptionPredictionsType          = "spot_interruption_predictions_type"
 	FieldNodeTemplateStorageOptimized                         = "storage_optimized"
+	FieldNodeTemplateStorageOptimizedEnum                     = "storage_optimized_enum"
```
linkas45, line 58: Suggest naming `FieldNodeTemplateStorageOptimizedStatus` as enum suffix doesn't add value and later on would be confusing

castai/resource_node_template.go
aweris: After some consideration, deprecating + conflict is a better approach here.  

Users getting clear warning about deprecation and what to do. 

```text
╷
│ Warning: Argument is deprecated
│
│   with castai_node_template.default_by_castai,
│   on castai.tf line 199, in resource "castai_node_template" "default_by_castai":
│  199:     storage_optimized  = false
│
│ Use `storage_optimized_enum` instead. Attribute will be removed in the next major release.
```

aweris: I just tried to remove the field and check the terraform changes. 

This is the only warning users get. There is no additional info. They need to go and look at documentation to understand what happened. 
```text
╷
│ Error: Unsupported argument
│
│   on castai.tf line 41, in resource "castai_node_template" "default_by_castai":
│   41:     storage_optimized = false
│
│ An argument named "storage_optimized" is not expected here.
```

castai/resource_node_template.go
```diff
@@ -1075,9 +1123,27 @@ func toTemplateConstraints(obj map[string]any) *sdk.NodetemplatesV1TemplateConst
...
-	if v, ok := obj[FieldNodeTemplateStorageOptimized].(bool); ok {
+
+	// Deprecated field, will be removed in the next major release.
+	if v, _ := obj[FieldNodeTemplateStorageOptimized].(bool); v {
 		out.StorageOptimized = toPtr(v)
 	}
+
+	// Until the next major release, we need to keep the backward compatibility with the old field set to true.
+	// otherwise, default value from the new field will be used and remove the constraint.
+	if !lo.FromPtr(out.StorageOptimized) {
```
aweris, line 1134: [note] This creates a side effect opposite of CSU.

`xx_optimized = false` is ignored and removed from the template even if it is set explicitly because we don't have a way to understand if it is set or zero value

However, this is also clearly indicated in the deprecation message.

castai/resource_node_template.go
linkas45: hmm, but if we still allow setting storage_optimized, then we have a problem with false value right? Because customer could ignore this warning and then get a side effect that their flag started not working

aweris: Yes, that's why the updated warning with explicit warning.  

I believe this migration is the most user-friendly approach if we're going to support it with proper change logs

```text
╷
│ Warning: Argument is deprecated
│
│   with castai_node_template.default_by_castai,
│   on castai.tf line 41, in resource "castai_node_template" "default_by_castai":
│   41:     storage_optimized = false
│
│ This attribute will be removed in the next major release. Use
│ `storage_optimized_status` instead. This field is currently only
│ processed if set to `true`.
│
│ (and one more similar warning elsewhere)
╵
```

castai/resource_node_template_test.go
```diff
@@ -459,6 +461,7 @@ func TestAccResourceNodeTemplate_basic(t *testing.T) {
 					resource.TestCheckResourceAttr(resourceName, "constraints.0.custom_priority.0.spot", "true"),
 					resource.TestCheckResourceAttr(resourceName, "constraints.0.custom_priority.0.on_demand", "true"),
 					resource.TestCheckResourceAttr(resourceName, "constraints.0.dedicated_node_affinity.#", "0"),
+					resource.TestCheckResourceAttr(resourceName, "constraints.0.storage_optimized_status", "disabled"),
```
linkas45, line 464: maybe also add check `resource.TestCheckResourceAttr(resourceName, "constraints.0. compute_optimized_status", ""),` ?

linkas45 (approved):

PR #311, opened 2024-05-09
Deprecate storage compute optimized bool

linkas45:
This marks `storage_optimized` and `compute_optimized` properties as deprecated, but their usage will still be allowed


saumas (approved):

castai/resource_node_template.go
```diff
@@ -54,6 +55,7 @@ const (
 	FieldNodeTemplateSpotInterruptionPredictionsEnabled       = "spot_interruption_predictions_enabled"
 	FieldNodeTemplateSpotInterruptionPredictionsType          = "spot_interruption_predictions_type"
 	FieldNodeTemplateStorageOptimized                         = "storage_optimized"
+	FieldNodeTemplateStorageOptimizedStatus                   = "storage_optimized_status"
```
furkhat, line 58: Could we please change the name? May be it is just me but I find that it sounds like a read-only attribute, not the value to set. Like in kubernetes status is not something to be set by user but only to be read.
```suggestion
	FieldNodeTemplateStorageOptimizedState                    = "storage_optimized_state"
```

castai/resource_node_template.go
```diff
@@ -221,8 +223,15 @@ func resourceNodeTemplate() *schema.Resource {
...
-							Description: "Storage optimized instance constraint - will only pick storage optimized nodes if true",
+							Description: "Storage optimized instance constraint - will only pick storage optimized nodes if true (deprecated).",
+							Deprecated:  "`storage_optimized` attribute will be removed in the next major release. Use `storage_optimized_status` instead.",
+						},
+						FieldNodeTemplateStorageOptimizedStatus: {
+							Type:             schema.TypeString,
+							Optional:         true,
+							Default:          "",
+							Description:      "Storage optimized instance constraint - will only pick storage optimized nodes if enabled and won't pick if disabled. Empty value will have no effect. Supported values: `enabled`, `disabled` or empty string.",
+							ValidateDiagFunc: validation.ToDiagFunc(validation.StringInSlice([]string{"", Enabled, Disabled}, false)),
```
furkhat, line 234: Did you have a chance to look at `ExactlyOnOf`? This would improve UX by displaying clear error if someone tries to configure both old and new fields at the same time.
```suggestion
							ValidateDiagFunc: validation.ToDiagFunc(validation.StringInSlice([]string{"", Enabled, Disabled}, false)),
							ExactlyOneOf: []string{fmt.Sprint(FieldNodeTemplateConstraints, ".0.", FieldNodeTemplateStorageOptimized), fmt.Print(FieldNodeTemplateConstraints, ".0.", FieldNodeTemplateStorageOptimizedStatus)},
```

Similar thing could be done for the other pair of fields.

furkhat (commented):
Left few suggestions.

castai/resource_node_template.go
linkas45: Hmm, do you have a proposal for this? I couldn't come up with a better name

castai/resource_node_template.go
linkas45: as we are defaulting to empty string, we will always get value from `FieldNodeTemplateStorageOptimizedStatus`, therefore this constraint will always fire if `FieldNodeTemplateStorageOptimized` is set

PR #315, opened 2024-05-13
Add deprecation message for storage and compute optimized constraints

linkas45:


castai/resource_node_template.go
```diff
@@ -223,6 +223,7 @@ func resourceNodeTemplate() *schema.Resource {
 							Optional:    true,
 							Default:     false,
 							Description: "Storage optimized instance constraint - will only pick storage optimized nodes if true",
+							Deprecated:  "Storage optimized constraint will be removed in the next major version and other resource for this functionality will have to be used.",
```
mikenorgate, line 226: This wording doesn't sound right, I feel it gives the impression we are going to be removing the constraint completely. Also we wont be replacing this with a different resource, just a different argument

linkas45: Hmm, agree, what do you think about this:
`Storage optimized constraint is deprecated and will be removed in the next major version. Other argument for the same functionality will be provided with the major version.`
Maybe you can think of other variations?

mikenorgate: `Storage optimized constraint is deprecated and will be replaced by a new argument in the next major release`?

linkas45: updated

furkhat (commented):
Could new fields be added along with deprecation?

Then deprecation message could state "Deprecated.. use xyz instead". And the next major release would just delete deprecated fields.

What do you think?

linkas45:
> Could new fields be added along with deprecation?
> 
> Then deprecation message could state "Deprecated.. use xyz instead". And the next major release would just delete deprecated fields.
> 
> What do you think?

Having 2 different properties changing single property will effect in terraform always seeing a difference when planning. If enum=true and boolean is not set (defaults to null). But then enum (because it's true) sets boolean value to true and then with another terraform plan there is a diff (true->null).

mikenorgate (approved):

PR #320, opened 2024-06-03
prefer actual error over context deadline exceeded

furkhat:


castai/sdk/api.gen.go
```diff
@@ -101,20 +101,6 @@ const (
 	STATUSUNKNOWN  CastaiSsoV1beta1SSOConnectionStatus = "STATUS_UNKNOWN"
 )
 
-// Defines values for ExternalclusterV1ClusterReconcileInfoReconcileMode.
```
Tsonov, line 104: Are the changes in this file expected, doesn't look related to the change?

castai/resource_aks_cluster.go
```diff
@@ -215,10 +216,15 @@ func updateAKSClusterSettings(ctx context.Context, data *schema.ResourceData, cl
...
 	// Retries are required for newly created IAM resources to initialise on Azure side.
 	b := backoff.WithContext(backoff.WithMaxRetries(backoff.NewConstantBackOff(10*time.Second), 30), ctx)
+	var lastErr error
 	if err = backoff.Retry(func() error {
 		response, err := client.ExternalClusterAPIUpdateClusterWithResponse(ctx, data.Id(), req)
+		lastErr = err
 		return sdk.CheckOKResponse(response, err)
 	}, b); err != nil {
+		if errors.Is(err, context.DeadlineExceeded) && lastErr != nil {
+			return fmt.Errorf("updating cluster configuration: %w", lastErr)
```
Tsonov, line 226: Maybe mention that the context deadline was exceeded and what the last observed error was (so we know it's also a timeout, would help debugging IMO)

You can either wrap both errors or just extend the error message here

castai/sdk/api.gen.go
furkhat: The PR fails because the SDK is not updated. It takes open api spec from the web so I had to re-generate it.

aldor007 (approved):

castai/resource_aks_cluster.go
furkhat: Done.

Tsonov (approved):

PR #334, opened 2024-06-13
validate examples and compatibility with TF modules using local changes

furkhat:


Makefile
```diff
@@ -1,14 +1,19 @@
+SHELL := /bin/bash
+
 default: build
 
 .PHONY: init-examples
 init-examples:
```
aldor007, line 6: I'm using this makefile target to test my local changes. Can you introduce a new one for GH action instead of modifying ?

furkhat: But you can test your changes now the same as before. Running `make init-examples` and then `terraform init` from within one of the examples will work.

aldor007 (approved):

PR #340, opened 2024-06-18
k8s version update

juliette-cast:
just updating the k8s version for AKS and EKS


juliette-cast (commented):
just changing the k8s version from 1.27 to 1.28 in AKS and EKS

juliette-cast:
I have added new commits for the Node.js update to 20 since that's why the checks failed .please review and merge thank you

PR #343, opened 2024-06-24
fix: filter null value from `gpu` constraint while getting diff

aweris:


furkhat (changes requested):
Please take a look if this much shorter approach achieves what you want https://github.com/castai/terraform-provider-castai/commit/dad895109b4f9b4d5e9c96267579a59f693e8bbc

aweris:
> Please take a look if this much shorter approach achieves what you want https://github.com/castai/terraform-provider-castai/commit/dad895109b4f9b4d5e9c96267579a59f693e8bbc

@furkhat, the issue is when user modules generate an empty `gpu` block, like the example above, which translates to the provider as `gpu: [null]` and we translate this as `gpu: null`. Everything is okay until this point. 

In the next execution, without any change in TF, the user still has `gpu: [null]` in his config but our API returns `gpu=null` which is diff for terraform, `[null] != []`, and shows a diff like the example above. 

This PR is explicitly suppressing diff when the user config is `[null]` and the API config is `null` or empty.

PR #357, opened 2024-08-07
feat: workload scaling policies

varnastadeus:


aivarasatkocaitiscastai (approved):

castai/resource_workload_scaling_policy.go
```diff
@@ -0,0 +1,341 @@
...
+		Schema: map[string]*schema.Schema{
+			"function": {
+				Type:             schema.TypeString,
+				Optional:         true,
+				Description:      "The function used to calculate the resource recommendation. Supported values: `QUANTILE`, `MAX`",
+				Default:          function,
+				ValidateDiagFunc: validation.ToDiagFunc(validation.StringInSlice([]string{"QUANTILE", "MAX"}, false)),
+			},
+			"args": {
+				Type:     schema.TypeList,
```
UOndro, line 98: maybe stupid question, but why is this not TypeFloat when it should be a single float number?

varnastadeus: I guess we had idea to support multiple args initially

castai/resource_workload_scaling_policy.go
```diff
@@ -0,0 +1,341 @@
...
+		ReadContext:   resourceWorkloadScalingPolicyRead,
+		UpdateContext: resourceWorkloadScalingPolicyUpdate,
+		DeleteContext: resourceWorkloadScalingPolicyDelete,
+		CustomizeDiff: resourceWorkloadScalingPolicyDiff,
+		Importer: &schema.ResourceImporter{
+			StateContext: workloadScalingPolicyImporter,
+		},
+		Description: "Manage workload scaling policy. Scaling policy [reference](https://docs.cast.ai/docs/woop-scaling-policies)",
+		Schema: map[string]*schema.Schema{
+			FieldClusterID: {
```
UOndro, line 36: nit: they could also set ConfidenceThreshold, but it is probably not important :D

castai/resource_workload_scaling_policy.go
```diff
@@ -0,0 +1,341 @@
...
+		return diag.FromErr(err)
+	}
+
+	delResp, err := client.WorkloadOptimizationAPIDeleteWorkloadScalingPolicyWithResponse(ctx, clusterID, d.Id())
+	if err != nil {
+		return diag.FromErr(err)
+	}
+	if delResp.StatusCode() == http.StatusBadRequest {
+		log.Printf("[WARN] Scaling policy has active workloads (%s) and can't be deleted, removing from state", d.Id())
+		return nil
```
UOndro, line 249: Shouldn't we return the error here too? Or maybe at least return Diagnostic{Severity: Warning, Summary:...}

castai/resource_workload_scaling_policy.go
```diff
@@ -0,0 +1,341 @@
...
+	}
```
UndeadRat22, line 250: I don't get this part, if we fail to delete it, we remove it from state?

UndeadRat22: I'm pretty sure we can get BadRequest not only for this err when deleting as well. Deleting readonly policy, deleting default policies 🤔.

castai/resource_workload_scaling_policy.go
```diff
@@ -0,0 +1,341 @@
...
+	if err := validateArgs(cpu, "cpu"); err != nil {
+		return err
+	}
+	return validateArgs(memory, "memory")
+}
+
+func validateArgs(r sdk.WorkloadoptimizationV1ResourcePolicies, res string) error {
+	if r.Function == "QUANTILE" && len(r.Args) == 0 {
+		return fmt.Errorf("field %q: QUANTILE function requires args to be provided", res)
+	}
```
UndeadRat22, line 272: validation will pass if len(Args) > 0 - so 2 args - ok.

castai/resource_workload_scaling_policy.go
```diff
@@ -0,0 +1,341 @@
...
+		return nil
+	}
+	if err := sdk.StatusOk(delResp); err != nil {
+		return diag.FromErr(err)
+	}
+
+	return nil
+}
+
+func resourceWorkloadScalingPolicyDiff(_ context.Context, d *schema.ResourceDiff, _ interface{}) error {
```
UndeadRat22, line 258: let's add some tests for custom validation

castai/resource_workload_scaling_policy.go
```diff
@@ -0,0 +1,341 @@
...
+
+	clusterID, id := ids[0], ids[1]
+	if err := d.Set(FieldClusterID, clusterID); err != nil {
+		return nil, fmt.Errorf("setting cluster id: %w", err)
+	}
+	d.SetId(id)
+
+	// Return if scaling policy ID provided.
+	if _, err := uuid.Parse(id); err == nil {
+		return []*schema.ResourceData{d}, nil
```
UOndro, line 293: shouldn't we check if it exists?

castai/resource_workload_scaling_policy.go
```diff
@@ -0,0 +1,341 @@
...
+	}
+
+	for _, sp := range resp.JSON200.Items {
+		if sp.Name == id {
+			d.SetId(sp.Id)
+			return []*schema.ResourceData{d}, nil
+		}
+	}
+
+	return nil, fmt.Errorf("failed to find workload scaling policy with the following name: %v", id)
```
UndeadRat22, line 310: wait there's something sus here, why is the `id` both a name and a uuid?

castai/resource_workload_scaling_policy.go
varnastadeus: there is validation on field itself with `MaxItems: 1`

castai/resource_workload_scaling_policy.go
varnastadeus: to have better ux and allow importing by providing name

castai/resource_workload_scaling_policy.go
varnastadeus: I added it to avoid unremovable state when eg you onboard and offboard your cluster, you would have to go to ui and unnassign all workloads just to remove from state which is annoying 🤔

castai/resource_workload_scaling_policy.go
varnastadeus: you mean scaling policy? it will go through read path and will return err if it doesn't exist

castai/resource_workload_scaling_policy.go
varnastadeus: same with default, it you import those you are stuck with them forever as you can't delete it so the only thing is to remove from state and let them be 🤔

varnastadeus: maybe for now I will remove that check for active workloads and add it if we get any complaints 🤔

castai/resource_workload_scaling_policy.go
```diff
@@ -0,0 +1,342 @@
...
+		Timeouts: &schema.ResourceTimeout{
+			Create: schema.DefaultTimeout(15 * time.Second),
+			Read:   schema.DefaultTimeout(15 * time.Second),
+			Update: schema.DefaultTimeout(15 * time.Second),
+			Delete: schema.DefaultTimeout(15 * time.Second),
+		},
+	}
+}
+
+func resourceSchema(function string, overhead float64) *schema.Resource {
```
furkhat, line 87: The current package defines all resources and schema therefore it is confusing what `resourceSchema` functions, which is callable from anywhere within the package means. Could you please rename it?

Note: you probably meant `resources` as in `cpu resource` but in the terraform provider's context the word "resource" hints a meaning like in `node_configurations resource`.

castai/resource_workload_scaling_policy.go
```diff
@@ -0,0 +1,342 @@
...
+	req := sdk.WorkloadOptimizationAPICreateWorkloadScalingPolicyJSONRequestBody{
+		Name:      d.Get("name").(string),
+		ApplyType: sdk.WorkloadoptimizationV1ApplyType(d.Get("apply_type").(string)),
+		RecommendationPolicies: sdk.WorkloadoptimizationV1RecommendationPolicies{
+			ManagementOption: sdk.WorkloadoptimizationV1ManagementOption(d.Get("management_option").(string)),
+		},
+	}
+
+	if v, ok := d.GetOk("cpu"); ok {
+		req.RecommendationPolicies.Cpu = toResourcePolicies(v.([]interface{})[0].(map[string]interface{}))
```
furkhat, line 139: Similar to what I mentioned in another comment. Could you please rename the function? Something that narrows the context.

castai/resource_workload_scaling_policy.go
```diff
@@ -0,0 +1,342 @@
...
+
+func resourceWorkloadScalingPolicyUpdate(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	if !d.HasChanges(
+		"name",
+		"apply_type",
+		"management_option",
+		"cpu",
+		"memory",
+	) {
+		log.Printf("[INFO] scaling policy up to date")
```
furkhat, line 203: Not sure why but in few other places we use `github.com/hashicorp/terraform-plugin-log/tflog` instead. I leave it up to you to evaluate if it makes sense to use it here.

castai/resource_workload_scaling_policy.go
```diff
@@ -0,0 +1,342 @@
...
+
+	client := meta.(*ProviderConfig).api
+	clusterID := d.Get(FieldClusterID).(string)
+	req := sdk.WorkloadOptimizationAPIUpdateWorkloadScalingPolicyJSONBody{
+		Name:      d.Get("name").(string),
+		ApplyType: sdk.WorkloadoptimizationV1ApplyType(d.Get("apply_type").(string)),
+		RecommendationPolicies: sdk.WorkloadoptimizationV1RecommendationPolicies{
+			ManagementOption: sdk.WorkloadoptimizationV1ManagementOption(d.Get("management_option").(string)),
+			Cpu:              toResourcePolicies(d.Get("cpu").([]interface{})[0].(map[string]interface{})),
+			Memory:           toResourcePolicies(d.Get("memory").([]interface{})[0].(map[string]interface{})),
```
furkhat, line 215: I didn't notice anything in `cpu` and `memory` fields that would ensure there is always an element there. Could there be a runtime error due to accessing zero index `[0]`?

castai/resource_workload_scaling_policy.go
```diff
@@ -0,0 +1,342 @@
...
+		out.Overhead = v
+	}
+	if v, ok := obj["apply_threshold"].(float64); ok {
+		out.ApplyThreshold = v
+	}
+
+	return out
+}
+
+func toResourceMap(p sdk.WorkloadoptimizationV1ResourcePolicies) []map[string]interface{} {
```
furkhat, line 333: Similar to other comments I would love to see a name with a more narrow meaning/context.

castai/resource_workload_scaling_policy_test.go
```diff
@@ -0,0 +1,181 @@
...
+					resource.TestCheckResourceAttr(resourceName, "cpu.0.function", "QUANTILE"),
+					resource.TestCheckResourceAttr(resourceName, "cpu.0.overhead", "0.15"),
+					resource.TestCheckResourceAttr(resourceName, "cpu.0.apply_threshold", "0.1"),
+					resource.TestCheckResourceAttr(resourceName, "cpu.0.args.0", "0.9"),
+					resource.TestCheckResourceAttr(resourceName, "memory.0.function", "QUANTILE"),
+					resource.TestCheckResourceAttr(resourceName, "memory.0.overhead", "0.35"),
+					resource.TestCheckResourceAttr(resourceName, "memory.0.apply_threshold", "0.2"),
+					resource.TestCheckResourceAttr(resourceName, "memory.0.args.0", "0.9"),
+				),
+			},
```
furkhat, line 60: Could you please add one more TestStep that verifying that Importer works correctly?

furkhat (changes requested):

castai/resource_workload_scaling_policy.go
UndeadRat22: well maybe in the code we can be explicit and be `nameOrID`

castai/resource_workload_scaling_policy.go
varnastadeus: there is `Required: true`,

furkhat (approved):

PR #365, opened 2024-08-19
fix: map SDK NodeAffinity operators to canonical provider values

vladklokun:
Before this commit, there was a difference
in how the SDK and the Terraform provider render the K8s Operator `In`. The SDK returned uppercase `IN`, while the Terraform provider expected `In`. This led to Terraform showing a difference in an infrastructure's state even when there was none.

This commit maps SDK Operator values to those defined in the provider.

### DIFF
--- a/castai/resource_node_template_test.go
+++ b/castai/resource_node_template_test.go
@@ -236,6 +236,56 @@ Tainted = false
 	)
 }
 
+func Test_flattenNodeAffinity(t *testing.T) {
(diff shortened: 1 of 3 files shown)


castai/resource_node_template_test.go
```diff
@@ -236,6 +236,56 @@ Tainted = false
 	)
 }
 
+func Test_flattenNodeAffinity(t *testing.T) {
```
linkas45, line 239: can you add a test which would test flattenNodeAffinity returning error?

vladklokun: Done.

PR #367, opened 2024-08-21
feat: Support CPU manufacturer node template constraint

mikenorgate:


castai/resource_organization_members.go
```diff
@@ -142,7 +143,7 @@ func resourceOrganizationMembersRead(ctx context.Context, data *schema.ResourceD
 	client := meta.(*ProviderConfig).api
 
 	organizationID := data.Id()
-	usersResp, err := client.UsersAPIListOrganizationUsersWithResponse(ctx, organizationID)
+	usersResp, err := client.UsersAPIListOrganizationUsersWithResponse(ctx, organizationID, &sdk.UsersAPIListOrganizationUsersParams{})
```
vladklokun, line 146: question: just to confirm, these are changes required by the newly generated SDK, right?

vladklokun (approved):
LGTM!

castai/resource_organization_members.go
mikenorgate: Yes

PR #379, opened 2024-09-23
CO-3070: feat: add pod pinner

vladklokun:
This commit adds Pod Pinner to the provider. Since Pod Pinner is enabled (`true`) by default, and we don't care if it is not set in Terraform, It is implemented as a boolean value instead of a tri-state toggle.


castai/resource_autoscaler.go
```diff
@@ -235,6 +236,22 @@ func resourceAutoscaler() *schema.Resource {
 											},
 										},
 									},
+									FieldPodPinner: {
+										Type:        schema.TypeList,
+										Optional:    true,
+										MaxItems:    1,
+										Description: "defines the Cast AI Pod Pinner components settings.",
```
aldor007, line 243:
```suggestion
										Description: "defines the Cast AI Pod Pinner components settings. Default enabled.",
```

aldor007 (approved):

castai/resource_autoscaler.go
vladklokun: Added default state to the `pod_pinner.enabled` field description.

PR #381, opened 2024-09-24
feat: add look back period to woop policy

jansyk13:


examples/resources/castai_workload_scaling_policy/resource.tf
```diff
@@ -1,13 +1,14 @@
...
   cpu {
-    function        = "QUANTILE"
-    overhead        = 0.15
-    apply_threshold = 0.1
-    args            = ["0.9"]
+    function                 = "QUANTILE"
+    overhead                 = 0.15
+    apply_threshold          = 0.1
+    args                     = ["0.9"]
+    look_back_period_seconds = 172800
```
JolantaSk, line 11: I guess this should be added to memory example too 🤔

JolantaSk (approved):

examples/resources/castai_workload_scaling_policy/resource.tf
jansyk13: This is example and the field is voluntary 🤷‍♂️

jansyk13:
```
cpu {
     apply_threshold          = 0.1
   }
``` 
vs
```
 memory {
     function        = "MAX"
    overhead        = 0.35
    apply_threshold = 0.2
  }
```

PR #386, opened 2024-09-25
KUBE-576: Add impersonate cluster support

stgleb:


castai/resource_gke_cluster_id.go
```diff
@@ -0,0 +1,188 @@
...
+	tkn, err := createClusterToken(ctx, client, clusterID)
+	if err != nil {
+		return diag.FromErr(err)
+	}
+	if err := data.Set(FieldClusterToken, tkn); err != nil {
+		return diag.FromErr(fmt.Errorf("setting cluster token: %w", err))
+	}
+	data.SetId(clusterID)
+	// If client service account is set, create service account on cast side.
+	if len(data.Get(FieldGKEClientSA).(string)) > 0 {
```
aldor007, line 124: Shouldn't this be a separate resource?

stgleb: it is actually a new resource https://github.com/castai/terraform-provider-castai/pull/386/files#diff-a5a684b5b5f8e9cbaac3f6a93c0e15e642c09317befcb3e7e11c0a6fbb557f6aR33 I use it to create registered cluster first and initialize it with cast service account and service key.

castai/provider.go
```diff
@@ -38,6 +38,7 @@ func Provider(version string) *schema.Provider {
 			"castai_eks_cluster":                resourceEKSCluster(),
 			"castai_eks_clusterid":              resourceEKSClusterID(),
 			"castai_gke_cluster":                resourceGKECluster(),
+			"castai_gke_cluster_id":             resourceGKEClusterId(),
```
furkhat, line 41:
```suggestion
			"castai_gke_clusterid":             resourceGKEClusterId(),
```
Just to match naming style for `castai_eks_clusterid`

furkhat: I actually don't understand, how is it different from `castai_gke_cluster`?

stgleb: I don't think this is crucial, also it will cause changes in tf code as well.

ValyaB (approved):

PR #388, opened 2024-09-30
eks_cluster_existing - Eliminating the error msg: "Schedule with a name \"rebalance spots at every 30th minute\" already exists"

marenascastai:
- Eliminating the error message bellow when you run a plan & apply: 
```
│ Error: expected status code 200, received: status=409 body={"message":"Schedule with a name \"rebalance spots at every 30th minute\" already exists", "fieldViolations":[]}
│ 
│   with castai_rebalancing_schedule.spots,
│   on castai.tf line 194, in resource "castai_rebalancing_schedule" "spots":
│  194: resource "castai_rebalancing_schedule" "spots" {
```

-> **count** 
Once implemented , you can conditionally control the number of instances of a resource to be created, usually based on a boolean variable.
Pros:
- Very flexible and directly tied to Terraform's built-in capabilities.
You can control the number of instances with exact precision.

Cons:
- If set to 0, the resource will not be created, but Terraform will remove it from the state file if it already exists.
__________________________________

-> **var.executions**
Pros:
- Useful if you want more control, such as toggling creation and destruction behavior with more logic.
- Good for handling multiple conditions, not just resource existence.

Cons:
- Requires more management if the logic is complex.
- Slightly less flexible for simple yes/no decisions compared to count.


marenascastai:
Will not be necessary as we can suggest the customer to comment the assign policy in their provisioning and just uncomment when appears a need

PR #392, opened 2024-10-01
feat: adding woop scaling policy min/max for cpu and memory

JolantaSk:


castai/resource_workload_scaling_policy.go
```diff
@@ -158,6 +159,17 @@ func workloadScalingPolicyResourceSchema(function string, overhead float64) *sch
 				Description:      "The look back period in seconds for the recommendation.",
 				ValidateDiagFunc: validation.ToDiagFunc(validation.IntBetween(24*60*60, 7*24*60*60)),
 			},
+			"min": {
+				Type:        schema.TypeFloat,
```
varnastadeus, line 163: not sure if we have validation in api, but does this value can be less than the default? 🤔

JolantaSk: It can, and it should probably not. Will add validation to API 🙏

varnastadeus (approved):

UndeadRat22 (approved):

PR #396, opened 2024-10-08
feat: add woop anti-affinity settings to tf scaling policy

UOndro:


JolantaSk (approved):

januszp-castai (approved):

castai/resource_workload_scaling_policy.go
```diff
@@ -125,6 +125,21 @@ func resourceWorkloadScalingPolicy() *schema.Resource {
 					},
 				},
 			},
+			"anti_affinity": {
```
aivarasatkocaitiscastai, line 128: Name is a bit misleading IMO. What does it mean to set anti-affinity to true or false? I think our API has a better name `consider_anti_affinity` maybe we can keep it in line with API naming?

aivarasatkocaitiscastai (approved):

PR #405, opened 2024-10-16
KUBE-364: added node_group_arn

cirisked:


castai/resource_node_configuration.go
```diff
@@ -286,6 +287,12 @@ func resourceNodeConfiguration() *schema.Resource {
 								},
 							},
 						},
+						"node_group_arn": {
```
Tsonov, line 290: Add a const for the field name (like the others); makes it easier to list all fields at the top

castai/resource_node_configuration.go
```diff
@@ -286,6 +287,12 @@ func resourceNodeConfiguration() *schema.Resource {
...
+							Type:             schema.TypeString,
+							Optional:         true,
+							Description:      "Cluster's node group ARN used for CAST provisioned node pools",
```
Tsonov, line 293: Is this only used for hibernate or for all CAST pools? Does it apply only to EKS?  Some additional context in the description would be useful.

cirisked: EKS only, will update description.

castai/resource_node_configuration.go
cirisked: done

castai/resource_node_configuration.go
```diff
@@ -178,6 +179,13 @@ func resourceNodeConfiguration() *schema.Resource {
 							},
 							Description: "Cluster's security groups configuration for CAST provisioned nodes",
 						},
+						"node_group_arn": {
```
aldor007, line 182: Can you also add acceptance test case?

cirisked: done

aldor007 (approved):

PR #423, opened 2024-11-15
CSU-2024: Drift detection for other clouds

Tsonov:
Similar logic to https://github.com/castai/terraform-provider-castai/pull/422 but for GKE and EKS. Unified some logic (mostly retry and drift) in one method. 

For EKS, there is no "drift" on Read because assumeRole is already exposed so we don't need detection - it is handled automatically. 

This means long-term (as we move to 100% impersonation on all clouds), the drift detection will not be needed.

### DIFF
--- a/castai/resource_gke_cluster.go
+++ b/castai/resource_gke_cluster.go
@@ -7,7 +7,6 @@ import (
 	"strings"
 	"time"
 
-	"github.com/cenkalti/backoff/v4"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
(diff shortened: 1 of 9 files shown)


castai/resource_gke_cluster.go
```diff
@@ -198,22 +205,8 @@ func updateGKEClusterSettings(ctx context.Context, data *schema.ResourceData, cl
...
-		if response.StatusCode() == 400 && !sdk.IsCredentialsError(response) {
-			return backoff.Permanent(err)
-		}
-		return err
-	}, backoff.NewExponentialBackOff()); err != nil {
-		return fmt.Errorf("updating cluster configuration: %w", err)
-	}
-
-	return nil
+	// todo backoff.NewExponentialBackOff()
```
furkhat, line 208: this can be deleted I think

furkhat (approved):
Thanks for thorough unit tests!

PR #424, opened 2024-11-17
Gitops to use token for access & general housekeeping

sarvesh-cast:
Summary:
1. TF Gitops approach was using API Key instead of API token, fixed it
2. Readme updated to detail gitops approach
3. Added clear distinction between API KEY & Cluster API token
4. General housekeeping

Where do we use API KEY vs API Token

1. Use API key only to run Terraform, but not pass it to Castware helm charts.
2. Use API token to in castware helm charts

**Explanation:**

- User API key is used by Terraform to makes calls to our API to provision our platform resources such as: onboarding cluster, creating/updating node configurations/templates, etc.

- To make cluster specific operations using castware API Token should be used.

- While it is possible to use the same API access key when configuring Castware components through TF it is an antipattern. And it may cause friction and downtime for customers when deleting such API keys.

### DIFF
--- a/examples/aks/aks_cluster_gitops/README.md
+++ b/examples/aks/aks_cluster_gitops/README.md
@@ -17,9 +17,17 @@ Helm Managed ==>  All Castware components such as `castai-agent`, `castai-cluste
                                                   2. Terraform Init & Apply| 
                                                 +-------------------------+
                                                             | 
+                                                            | 
+                                                            | TERRAFORM OUTPUT
+                                                +-------------------------+
+                                                |  3. Execute terraform output command
+                                                | terraform output cluster_id  
+                                                  terraform output cluster_token
+                                                +-------------------------+
+                                                            | 
                                                             |GITOPS
                                                 +-------------------------+
-                                                | 3. Deploy Helm chart of castai-agent castai-cluster-controller`, `castai-evictor`, `castai-spot-handler`, `castai-kvisor`, `castai-workload-autoscaler`, `castai-pod-pinner`
+                                                | 4. Deploy Helm chart of castai-agent castai-cluster-controller`, `castai-evictor`, `castai-spot-handler`, `castai-kvisor`, `castai-workload-autoscaler`, `castai-pod-pinner`
                                                 +-------------------------+         
                                                             | 
                                                             | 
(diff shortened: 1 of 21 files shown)


examples/aks/aks_cluster_gitops/README.md
```diff
@@ -34,28 +42,32 @@ Prerequisites:
...
 
 ### Step 3: Deploy Helm chart of CAST Components
 Coponents: `castai-cluster-controller`,`castai-evictor`, `castai-spot-handler`, `castai-kvisor`, `castai-workload-autoscaler`, `castai-pod-pinner` \
 After all CAST AI components are installed in the cluster its status in CAST AI console would change from `Connecting` to `Connected` which means that cluster onboarding process completed successfully.
 
 ```
-CASTAI_API_KEY=""
-CASTAI_CLUSTER_ID=""
+CASTAI_CLUSTER_ID="<Replace cluster_id>"
+CASTAI_CLUSTER_API_TOKEN="<Replace cluster_token>"
```
andrejatcastai, line 59: May be it's better to use _KEY here as well?

examples/aks/aks_cluster_gitops/README.md
```diff
@@ -104,4 +116,10 @@ helm upgrade -i castai-kvisor castai-helm/castai-kvisor -n castai-agent \
 
 This example can also be used to import AKS cluster to Terraform which is already onboarded to CAST AI console through [script](https://docs.cast.ai/docs/cluster-onboarding#how-it-works).   
 For importing existing cluster follow steps 1-3 above and change `castai_node_configuration.default` Node Configuration name.
-This would allow to manage already onboarded clusters' CAST AI Node Configurations and Node Templates through IaC.
\ No newline at end of file
+This would allow to manage already onboarded clusters' CAST AI Node Configurations and Node Templates through IaC.
+
+## Note: Difference between CAST API key & CAST Cluster API Token
```
andrejatcastai, line 121: I am not sure this is the case. Could you elaborate?

examples/eks/eks_cluster_gitops/outputs.tf
```diff
@@ -3,6 +3,12 @@ output "cluster_id" {
   description = "CAST AI cluster ID"
 }
 
+output "cluster_token" {
```
andrejatcastai, line 6: API Key is very sensitive, why do you suggest to expose it here? It is provided as an input, hence it's available anyway ;)

examples/aks/aks_cluster_gitops/README.md
sarvesh-cast: TF flow you should use user API access key only to run Terraform, but not pass it to Castware helm charts. 
It means that using your user API access key Terraform makes calls to our API to provision our platform resources such as: onboarding cluster, creating/updating node configurations/templates, etc.

To make cluster specific operations using castware API Token should be used.

While it is possible to use the same API access key when configuring Castware components through TF it is an antipattern. And it may cause friction and downtime for customers when deleting such API keys.

examples/aks/aks_cluster_gitops/README.md
sarvesh-cast: https://github.com/castai/terraform-provider-castai/pull/424#discussion_r1846563344

examples/eks/eks_cluster_gitops/outputs.tf
sarvesh-cast: Yes, have passed a flag as sensitive = true. This will not display the api token.

Why do we need this ?

we need to run terraform output cluster_token to get the output and pass to castware helmcharts

PR #429, opened 2024-11-29
WIRE-837 - Support for managing role bindings

oskarwojciski:
Add support for managing role bidings (only for organizations with rbacv2 enabled)


castai/resource_role_bindings.go
```diff
@@ -0,0 +1,408 @@
...
+			Description: lo.ToPtr(data.Get(FieldRoleBindingsDescription).(string)),
+			Name:        data.Get(FieldRoleBindingsName).(string),
+		},
+	})
+
+	if err := sdk.CheckOKResponse(resp, err); err != nil {
+		return diag.FromErr(fmt.Errorf("create role binding: %w", err))
+	}
+
+	if len(*resp.JSON200) == 0 {
```
radekska, line 207: would we ever get to this point? `sdk.CheckOKResponse` isn't gonna handle the 5XX / 4XX cases?

castai/resource_role_bindings_test.go
```diff
@@ -0,0 +1,112 @@
...
+	"github.com/golang/mock/gomock"
+	"github.com/hashicorp/go-cty/cty"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/terraform"
+	"github.com/stretchr/testify/require"
+
+	"github.com/castai/terraform-provider-castai/castai/sdk"
+	mock_sdk "github.com/castai/terraform-provider-castai/castai/sdk/mock"
+)
+
+func TestRoleBindingsReadContext(t *testing.T) {
```
radekska, line 19: I'd add tests for other contexts as well

radekska (approved):

castai/resource_role_bindings.go
oskarwojciski: Theoretically not. I wanted to be double sure that we don't run into `index out of range` in line 211

PR #430, opened 2024-12-01
WIRE-839 add service account resource

radekska:


castai/resource_service_account.go
```diff
@@ -0,0 +1,175 @@
...
+			FieldServiceAccountName: {
+				Type:        schema.TypeString,
+				Required:    true,
+				ForceNew:    true,
+				Description: "Name of the service account.",
+			},
+			FieldServiceAccountDescription: {
+				Type:        schema.TypeString,
+				Optional:    true,
+				ForceNew:    true,
```
oskarwojciski, line 56: Should the resource be recreated on Name or Description change?
We do not allow to update the Service Account - is it on purpose?

radekska: We didn't create the endpoint for it as there was no need from the UI perspective

radekska: But I'm gonna add the method as it doesn't make sense to recreate resources just because of name or description.

radekska: Updated

castai/resource_service_account.go
```diff
@@ -0,0 +1,250 @@
...
+	if err := data.Set(FieldServiceAccountAuthor, authorData); err != nil {
+		return diag.Errorf("setting service account author: %v", err)
+	}
+	return nil
+}
+
+func resourceServiceAccountCreate(ctx context.Context, data *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	client := meta.(*ProviderConfig).api
+
+	organizationID := data.Get(FieldServiceAccountOrganizationID).(string)
```
oskarwojciski, line 135: There is a way to get the default organization from the token, it can be used when the organization is empty:
```
organizationID, err = getDefaultOrganizationId(ctx, meta)
```

examples/resources/castai_service_account/resource.tf
```diff
@@ -0,0 +1,5 @@
+resource "castai_service_account" "service_account" {
+  organization_id = organization.id
+  name            = "service-account-name"
+  description     = "service account description"
+}
```
oskarwojciski, line 5: What about adding an email and author to the examples?

radekska: I am not sure as those are read-only fields, adding them would indicate to users that they can set them

castai/resource_service_account.go
radekska: updated

examples/resources/castai_service_account/resource.tf
oskarwojciski: Oh, I thought we can set the service account email. so that is ok

oskarwojciski (approved):

PR #431, opened 2024-12-10
WIRE-839: add service account keys support

radekska:
- resource implementation for `castai_service_account_key` 
- minor tweaks for `castai_service_account` resource


castai/resource_service_account.go
```diff
@@ -278,3 +295,276 @@ func getOrganizationID(ctx context.Context, data *schema.ResourceData, meta inte
 
 	return organizationID, nil
 }
+
+func resourceServiceAccountKey() *schema.Resource {
```
oskarwojciski, line 299: If we treat this as separate resource, why not make a separate file `resource_service_account_key.go` to keep the files smaller?

oskarwojciski (commented):
Can you add some example of working with keys, also how to obtain the key token value while creating it?

I wonder if we want to split the resources or make the keys part of the `resource_service_account`. They are separated in the API, but logically, they are very much connected.

castai/resource_service_account.go
```diff
@@ -0,0 +1,570 @@
...
+				Required:         true,
+				Description:      "Name of the service account.",
+				ValidateDiagFunc: validation.ToDiagFunc(validation.StringIsNotWhiteSpace),
+			},
+			FieldServiceAccountDescription: {
+				Type:        schema.TypeString,
+				Optional:    true,
+				Description: "Description of the service account.",
+			},
+			FieldServiceAccountEmail: {
```
Vizualni, line 74: Do we also need to provide the service account id field?

radekska:
> Can you add some example of working with keys, also how to obtain the key token value while creating it?
>
Examples added

> I wonder if we want to split the resources or make the keys part of the `resource_service_account`. They are separated in the API, but logically, they are very much connected.

I thought about the same and decided to follow GCP approach:
- https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/google_service_account
- https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/google_service_account_key

oskarwojciski:
I see that we show change state for the name change, but we don't allow to update it, so it never apply
image
should we add `ForceNew` flag for `name`, and the `service_account_id` fields? Same as for `expires_at` which works fine

castai/resource_service_account.go
radekska: Could you elaborate on that? I'm not sure what you meant

PR #433, opened 2024-12-16
Add support for setting cpu&mem limit strategy

mszostok:
**Description**

- Add support for setting cpu&mem limit strategy

## Demo

https://github.com/user-attachments/assets/bc74b973-c5d9-4e12-b357-a157703346a7

### DIFF
--- a/castai/resource_workload_scaling_policy.go
+++ b/castai/resource_workload_scaling_policy.go
@@ -231,10 +231,7 @@ func workloadScalingPolicyResourceLimitSchema() *schema.Resource {
 				Description: fmt.Sprintf(`Defines limit strategy type.
 	- %s - removes the resource limit even if it was specified in the workload spec.
 	- %s - used to calculate the resource limit. The final value is determined by multiplying the resource request by the specified factor.`, sdk.NOLIMIT, sdk.MULTIPLIER),
-				ValidateDiagFunc: validation.ToDiagFunc(validation.StringInSlice([]string{string(sdk.MULTIPLIER), string(sdk.NOLIMIT)}, true)),
-				DiffSuppressFunc: func(k, oldValue, newValue string, d *schema.ResourceData) bool {
-					return strings.EqualFold(oldValue, newValue)
-				},
+				ValidateDiagFunc: validation.ToDiagFunc(validation.StringInSlice([]string{string(sdk.MULTIPLIER), string(sdk.NOLIMIT)}, false)),
 			},
 			FieldLimitStrategyMultiplier: {
 				Type:             schema.TypeFloat,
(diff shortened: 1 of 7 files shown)


castai/resource_workload_scaling_policy.go
```diff
@@ -203,6 +211,37 @@ func workloadScalingPolicyResourceSchema(function string, overhead, minRecommend
...
+}
+
+func workloadScalingPolicyResourceLimitSchema() *schema.Resource {
+	return &schema.Resource{
+		Schema: map[string]*schema.Schema{
+			FieldLimitStrategyType: {
+				Type:     schema.TypeString,
+				Required: true,
+				Description: `Defines limit strategy type.
+	- NONE - removes the resource limit even if it was specified in the workload spec.
```
aivarasatkocaitiscastai, line 232: Didn't we replace `none` with `no limit`? 🤔

castai/resource_workload_scaling_policy.go
```diff
@@ -449,10 +513,28 @@ func toWorkloadScalingPolicies(obj map[string]interface{}) sdk.Workloadoptimizat
 	if v, ok := obj["max"].(float64); ok && v > 0 {
 		out.Max = lo.ToPtr(v)
 	}
+	if v, ok := obj["limit"].([]any); ok && len(v) > 0 {
```
aivarasatkocaitiscastai, line 516: `FieldLimitStrategy` vs `"limit"`

aivarasatkocaitiscastai: I suppose here prior map access was with a const but now we have a nicely named one

castai/resource_workload_scaling_policy.go
```diff
@@ -203,6 +211,37 @@ func workloadScalingPolicyResourceSchema(function string, overhead, minRecommend
...
+	return &schema.Resource{
+		Schema: map[string]*schema.Schema{
+			FieldLimitStrategyType: {
+				Type:     schema.TypeString,
+				Required: true,
+				Description: `Defines limit strategy type.
+	- NONE - removes the resource limit even if it was specified in the workload spec.
+	- MULTIPLIER - used to calculate the resource limit. The final value is determined by multiplying the resource request by the specified factor.`,
+				ValidateDiagFunc: validation.ToDiagFunc(validation.StringInSlice([]string{string(sdk.MULTIPLIER), string(sdk.NOLIMIT)}, true)), // FIXME: any reason to be strict about it?
+				DiffSuppressFunc: func(k, oldValue, newValue string, d *schema.ResourceData) bool { // FIXME:
```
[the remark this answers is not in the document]
mszostok: actually I've added this comment to make sure that I will ask about that part of the code. 
In other components, I saw that they are ignoring casing, and I agree with this approach.

However, in the policy schema, other props, e.g. `apply_type`, are case-sensitive. Is there a reason for this? I'm torn between having consistent approach and making it less error-prone 😅

castai/resource_workload_scaling_policy.go
mszostok: fixed 👍

castai/resource_workload_scaling_policy.go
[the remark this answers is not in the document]
aivarasatkocaitiscastai: At this point I'd go with what's similar with managing the policy to not introduce more variations. Other than that I'd follow how bigger terraform providers handle such cases

aivarasatkocaitiscastai (approved):

PR #438, opened 2025-01-14
feat: add woop apply threshold strategy to tf scaling policy

UOndro:


castai/resource_workload_scaling_policy.go
```diff
@@ -251,6 +269,27 @@ func workloadScalingPolicyResourceLimitSchema() *schema.Resource {
 	}
 }
 
+func workloadScalingPolicyResourceApplyThresholdStrategySchema() *schema.Resource {
```
mszostok, line 272: ok, so the most problematic part is the proper resource mapping. Running `terraform apply -var-file=tf.vars` twice without changing anything result in invalid diff:
1. default `apply_threshold` is put to default `0.1` if not specified, but API returns the one that was populated by strategy.
    ```tf
      ~ resource "castai_workload_scaling_policy" "woop-541-e2e-test" {
            id                = "6ac1821a-36a8-4bbd-9d74-49b8a9808aae"
            name              = "woop-541-e2e-test"
            # (3 unchanged attributes hidden)
    
          ~ memory {
              ~ apply_threshold          = 0.2 -> 0.1
                # (6 unchanged attributes hidden)
    
                # (1 unchanged block hidden)
            }
    
            # (5 unchanged blocks hidden)
        }
    ```
2. when I use the  `apply_threshold` or don't specify it (use default) then API returns corresponding new strategy, but it's missing from TF definition, so it wants to remove it:
   
    ```tf
      ~ resource "castai_workload_scaling_policy" "woop-541-e2e-test" {
            id                = "423635de-1a54-4fc6-9a10-6039a273e76c"
            name              = "woop-541-e2e-test"
            # (3 unchanged attributes hidden)
    
          ~ cpu {
                # (7 unchanged attributes hidden)
    
              - apply_threshold_strategy {
                  - percentage = 0.1 -> null
                  - type       = "PERCENTAGE" -> null
                }
    
                # (1 unchanged block hidden)
            }
    
          ~ memory {
                # (7 unchanged attributes hidden)
    
              - apply_threshold_strategy {
                  - percentage = 0.2 -> null
                  - type       = "PERCENTAGE" -> null
                }
            }
    
            # (4 unchanged blocks hidden)
        }
    ```

castai/resource_workload_scaling_policy.go
```diff
@@ -3,6 +3,7 @@ package castai
 import (
 	"context"
 	"fmt"
+	"github.com/pkg/errors"
```
UndeadRat22, line 6: is this import used anywhere else in the provider code? Why is the built in errors pkg not sufficient? 🤔

castai/resource_workload_scaling_policy.go
```diff
@@ -437,6 +476,32 @@ func validateResourcePolicy(r sdk.WorkloadoptimizationV1ResourcePolicies, res st
...
+}
+
+func validatePercentageThresholdStrategy(r *sdk.WorkloadoptimizationV1ApplyThresholdStrategyPercentageThreshold) error {
+	if r.Percentage < minApplyThresholdValue || r.Percentage > maxApplyThresholdValue {
+		return fmt.Errorf(
+			`field %q: value must be between %v and %v`,
+			FieldApplyThresholdStrategyPercentage, minApplyThresholdValue, maxApplyThresholdValue)
+	}
 	return nil
 }
```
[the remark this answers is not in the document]
UOndro: Changed it to check only if the value is 0. If the value is 0, it means it was not specified. 

The percentage is optional and you can do something like this:
```
    apply_threshold_strategy {
      type        = "PERCENTAGE"
    }
```

castai/resource_workload_scaling_policy.go
UOndro: Fk hell, this was such a pain to fix it. I tired multiple things, and finally I found something this in [documentation](https://developer.hashicorp.com/terraform/plugin/sdkv2/best-practices/deprecations#renaming-an-optional-attribute).

In summary, I looked at what is set in `ResourceData` and decided to set `apply_threshold` or `apply_threshold_strategy`. The only issue is that during import, there is no previous value, so I decided to set only `apply_threshold_strategy`. Then, if the client uses `apply_threshold`, it will act as if there is a change, although it might be the same. After applying, it should be fixed, and reapplying will show no change.

Not much experience with terraform, is it ok?

castai/resource_workload_scaling_policy.go
UOndro: Old habit, but built-in pkg has no Wrap function. I didn't see it anywhere else, so I just reworked it

castai/resource_workload_scaling_policy.go
```diff
@@ -340,6 +379,13 @@ func resourceWorkloadScalingPolicyRead(ctx context.Context, d *schema.ResourceDa
...
 
+func getResourceFrom(d *schema.ResourceData, resource string) map[string]interface{} {
+	if v, ok := d.GetOk(resource); ok {
+		return v.([]interface{})[0].(map[string]interface{})
+	}
+	return map[string]interface{}{}
+}
```
mszostok, line 387: **nit**
```suggestion
func getResourceFrom(d *schema.ResourceData, resource string) map[string]any {
	if v, ok := d.GetOk(resource); ok {
		return v.([]any)[0].(map[string]any)
	}
	return map[string]any{}
}
```

castai/resource_workload_scaling_policy.go
```diff
@@ -437,6 +482,38 @@ func validateResourcePolicy(r sdk.WorkloadoptimizationV1ResourcePolicies, res st
...
+
+func validatePercentageThresholdStrategy(r *sdk.WorkloadoptimizationV1ApplyThresholdStrategyPercentageThreshold) error {
+	if r.Percentage == 0 {
+		return fmt.Errorf(
+			`field %q: field %q: value must be set for strategy type %s`,
+			FieldApplyThresholdStrategy,
+			FieldApplyThresholdStrategyPercentage,
+			FieldApplyThresholdStrategyPercentageType,
+		)
```
mszostok, line 515:
```suggestion
		return fmt.Errorf(
			`field %q: value must be set for strategy type %s`,
			FieldApplyThresholdStrategyPercentage,
			FieldApplyThresholdStrategyPercentageType,
		)
```

it's handled in the above layer, so not it's duplicated: `field "cpu": field "apply_threshold_strategy": field "apply_threshold_strategy": field "percentage": value must be set for strategy type PERCENTAGE`

castai/resource_workload_scaling_policy.go
```diff
@@ -524,10 +598,26 @@ func toWorkloadScalingPolicies(obj map[string]interface{}) sdk.Workloadoptimizat
 	if v, ok := obj["management_option"].(string); ok && v != "" {
 		out.ManagementOption = lo.ToPtr(sdk.WorkloadoptimizationV1ManagementOption(v))
 	}
+	out.ApplyThresholdStrategy = resolveApplyThresholdStrategy(obj)
 
 	return out
 }
 
+func resolveApplyThresholdStrategy(obj map[string]interface{}) *sdk.WorkloadoptimizationV1ApplyThresholdStrategy {
+	if v, ok := obj["apply_threshold"].(float64); ok && v > 0 {
```
mszostok, line 607: **nit** WDYT about putting that into sth like `const DeprecatedFieldApplyThreshold = "apply_threshold"`, because if I got it correctly is the deprecated option? In such situation, doing that conversion from old to new approach makes a lot of sens to me

castai/resource_workload_scaling_policy.go
```diff
@@ -567,13 +680,37 @@ func toWorkloadScalingPoliciesMap(p sdk.WorkloadoptimizationV1ResourcePolicies)
 		m[FieldLimitStrategy] = []map[string]any{limit}
 	}
 
+	if v, ok := previousCfg["apply_threshold"].(float64); ok && v > 0 {
+		m["apply_threshold"] = p.ApplyThreshold
+	} else {
+		strategy := applyThresholdStrategyToMap(*p.ApplyThresholdStrategy)
```
mszostok, line 686: should we check for nil here 🤔 if I get it correctly, BE always return it, but in case of bug, maybe it's better to not panic here

castai/resource_workload_scaling_policy.go
```diff
@@ -567,13 +680,37 @@ func toWorkloadScalingPoliciesMap(p sdk.WorkloadoptimizationV1ResourcePolicies)
 		m[FieldLimitStrategy] = []map[string]any{limit}
 	}
 
+	if v, ok := previousCfg["apply_threshold"].(float64); ok && v > 0 {
```
mszostok, line 683: **nit** I would extract that into a func similar to what you did with `resolveApplyThresholdStrategy`. Then, place both functions close to each other and include a "business" comment explaining why it’s done this way, or just add a link to a gh comments, so even after N months, we’ll still be able to quickly understand “what the hell..” 😄

castai/resource_workload_scaling_policy.go
mszostok: > Not much experience with terraform, is it ok?

Unfortunately, I'm not that experience with TF best practices, but for me it's ok.

However, there is one more case that is failing now:

```tf
provider "castai" {
  api_url   = var.castai_api_url
  api_token = var.castai_api_token
}

resource "castai_workload_scaling_policy" "woop-541-e2e-test" {
	name              = "woop-541-e2e-test"
	cluster_id        = var.castai_gke_cluster_id
	management_option = "MANAGED"
	cpu {
		min                      = 0.1
		max                      = 1
		limit {
			type       = "MULTIPLIER"
			multiplier = 1.2
		}
	}
	memory {
		function        = "MAX"
		overhead        = 0.35
	}
	startup {
		period_seconds = 240
	}
	downscaling {
		apply_type = "DEFERRED"
	}
	memory_event {
		apply_type = "IMMEDIATE"
	}
	anti_affinity {
	}
}
```

running apply twice results in diff:
```shell
  # castai_workload_scaling_policy.woop-541-e2e-test will be updated in-place
  ~ resource "castai_workload_scaling_policy" "woop-541-e2e-test" {
        id                = "dddb7094-9856-421b-b1f0-21906fd34ee8"
        name              = "woop-541-e2e-test1"
        # (3 unchanged attributes hidden)

      ~ cpu {
            # (7 unchanged attributes hidden)

          - apply_threshold_strategy {
              - percentage = 0.1 -> null
              - type       = "PERCENTAGE" -> null
            }

            # (1 unchanged block hidden)
        }

      ~ memory {
            # (7 unchanged attributes hidden)

          - apply_threshold_strategy {
              - percentage = 0.1 -> null
              - type       = "PERCENTAGE" -> null
            }
        }

        # (4 unchanged blocks hidden)
    }

Plan: 0 to add, 1 to change, 0 to destroy.

```

tested on `master` and we don't have such problem there with the old `apply_threshold` prop, it can be truly optional.

mszostok (approved):
Tested all the different edge cases, and it worked like a charm 😄

italian-si

castai/resource_workload_scaling_policy.go
```diff
@@ -3,6 +3,7 @@ package castai
 import (
 	"context"
 	"fmt"
+	"maps"
```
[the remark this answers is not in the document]
UOndro: ahhhh, that's why building fails

PR #443, opened 2025-01-21
Handle the creation of default scaling policies

mszostok:
**Description**

This PR allows to manage default scaling policies without importing them. If there is create operation against the default scaling policy, it's changed to update.

Without this PR, the following error is thrown:
```tf
   ╷
   │ Error: expected status code 200, received: status=409 body={"message":"scaling policy with the same name already exists", "fieldViolations":[]}
   │
   │   with castai_workload_scaling_policy.default["staging"],
   │   on import.tf line 8, in resource "castai_workload_scaling_policy" "default":
   │    8: resource "castai_workload_scaling_policy" "default" {
   ╵
```

I also update the import section to describe easier way of doing that.

**Limitations**
1. Default policies are created in the context of a given cluster ID. As a result, you need to collect and provide all the cluster IDs where you want to change the default policy.
2. Most of the props are required anyway so if they want to update only a few props, they are still required to provide all of them.

### DIFF
--- a/castai/resource_workload_scaling_policy.go
+++ b/castai/resource_workload_scaling_policy.go
@@ -323,14 +323,32 @@ func resourceWorkloadScalingPolicyCreate(ctx context.Context, d *schema.Resource
 
 	req.RecommendationPolicies.AntiAffinity = toAntiAffinity(toSection(d, "anti_affinity"))
 
-	resp, err := client.WorkloadOptimizationAPICreateWorkloadScalingPolicyWithResponse(ctx, clusterID, req)
-	if checkErr := sdk.CheckOKResponse(resp, err); checkErr != nil {
-		return diag.FromErr(checkErr)
+	create, err := client.WorkloadOptimizationAPICreateWorkloadScalingPolicyWithResponse(ctx, clusterID, req)
+	if err != nil {
+		return diag.FromErr(err)
 	}
 
-	d.SetId(resp.JSON200.Id)
+	switch create.StatusCode() {
+	case http.StatusOK:
+		d.SetId(create.JSON200.Id)
+		return resourceWorkloadScalingPolicyRead(ctx, d, meta)
+	case http.StatusConflict:
+		list, err := client.WorkloadOptimizationAPIListWorkloadScalingPoliciesWithResponse(ctx, clusterID)
+		if checkErr := sdk.CheckOKResponse(list, err); checkErr != nil {
+			return diag.FromErr(checkErr)
+		}
 
-	return resourceWorkloadScalingPolicyRead(ctx, d, meta)
+		for _, sp := range list.JSON200.Items {
+			if sp.Name == req.Name && sp.IsDefault {
+				d.SetId(sp.Id)
+				return resourceWorkloadScalingPolicyUpdate(ctx, d, meta)
+			}
+		}
+		return diag.Errorf("scaling policy with name %q already exists", req.Name)
+	default:
+		return diag.Errorf("expected status code %d, received: status=%d body=%s", http.StatusOK, create.StatusCode(), string(create.GetBody()))
+	}
+
 }
 
 func resourceWorkloadScalingPolicyRead(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
@@ -542,12 +560,11 @@ func validateResourceLimit(r sdk.WorkloadoptimizationV1ResourcePolicies) error {
 }
 
 func workloadScalingPolicyImporter(ctx context.Context, d *schema.ResourceData, meta interface{}) ([]*schema.ResourceData, error) {
-	ids := strings.Split(d.Id(), "/")
-	if len(ids) != 2 || ids[0] == "" || ids[1] == "" {
+	clusterID, nameOrID, found := strings.Cut(d.Id(), "/")
+	if !found {
 		return nil, fmt.Errorf("expected import id with format: <cluster_id>/<scaling_policy name or id>, got: %q", d.Id())
 	}
 
-	clusterID, nameOrID := ids[0], ids[1]
 	if err := d.Set(FieldClusterID, clusterID); err != nil {
 		return nil, fmt.Errorf("setting cluster nameOrID: %w", err)
 	}
--- a/castai/resource_workload_scaling_policy.go
+++ b/castai/resource_workload_scaling_policy.go
@@ -333,22 +333,18 @@ func resourceWorkloadScalingPolicyCreate(ctx context.Context, d *schema.Resource
 		d.SetId(create.JSON200.Id)
 		return resourceWorkloadScalingPolicyRead(ctx, d, meta)
 	case http.StatusConflict:
-		list, err := client.WorkloadOptimizationAPIListWorkloadScalingPoliciesWithResponse(ctx, clusterID)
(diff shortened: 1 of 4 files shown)


templates/resources/workload_scaling_policy.md.tmpl
```diff
@@ -18,24 +18,105 @@ simultaneously or create custom policies with different settings and apply them
 
 {{ .SchemaMarkdown | trimspace }}
 
-
 ## Importing
-You can use the `terraform import` command to import existing scaling policy to Terraform state.
```
UOndro, line 23: I like that we added there how to do it with import block, but I would also keep this section, maybe put it at the bottom and write here that this is the preferred way

castai/resource_workload_scaling_policy.go
```diff
@@ -279,14 +279,32 @@ func resourceWorkloadScalingPolicyCreate(ctx context.Context, d *schema.Resource
...
+		return diag.FromErr(err)
 	}
 
-	d.SetId(resp.JSON200.Id)
+	switch create.StatusCode() {
+	case http.StatusOK:
+		d.SetId(create.JSON200.Id)
+		return resourceWorkloadScalingPolicyRead(ctx, d, meta)
+	case http.StatusConflict:
+		list, err := client.WorkloadOptimizationAPIListWorkloadScalingPoliciesWithResponse(ctx, clusterID)
```
UOndro, line 292: I would wrap this in a function and use it also in import, it could return policy and then we can check if it is default

templates/resources/workload_scaling_policy.md.tmpl
aivarasatkocaitiscastai: Terraform import block is only supported from certain version 🙏

mszostok: yes, since 1.5.0 (June 12, 2023) so quite a long time ago. But I will put back the old description too 👍

castai/resource_workload_scaling_policy.go
mszostok: but import is fully read only, don't mam any other props, only id, so in the upper layer I won't have the info whether is default.  Unifying that IMO will make it less readable 🤔 

If possible pls provide some pseudo-code that I can follow 👍

templates/resources/workload_scaling_policy.md.tmpl
aivarasatkocaitiscastai: Enterprises are slow to update 😄 Had first hand experience

castai/resource_workload_scaling_policy.go
UOndro:
```go
func getPolicyByName(ctx context.Context, clusterID, name string, client sdk.ClientWithResponsesInterface) (*sdk.WorkloadoptimizationV1WorkloadScalingPolicy, error) {
	list, err := client.WorkloadOptimizationAPIListWorkloadScalingPoliciesWithResponse(ctx, clusterID)
	if checkErr := sdk.CheckOKResponse(list, err); checkErr != nil {
		return nil, checkErr
	}

	for _, sp := range list.JSON200.Items {
		if sp.Name == name {
			return &sp, nil
		}
	}
	return nil, fmt.Errorf("did not found policy with name %q", name)
}
```

and then:
```
		sp, err := getPolicyByName(ctx, clusterID, req.Name, client)
		if err != nil {
			return diag.FromErr(err)
		}
		if sp.IsDefault {
			return resourceWorkloadScalingPolicyUpdate(ctx, d, meta)
		}
```

mszostok: For some reason, I thought you wanted to have a generic workloadScalingPolicyImporter and use it in both places… Thanks for sharing the snippet; it helped clear my brain fart 😄

mszostok:
bumping the SDK causes tests to fail, I will report that to the relevant teams.

templates/resources/workload_scaling_policy.md.tmpl
```diff
@@ -18,12 +18,111 @@ simultaneously or create custom policies with different settings and apply them
 
 {{ .SchemaMarkdown | trimspace }}
 
-
 ## Importing
+
+For each connected cluster, a default scaling policy is created. An existing scaling policy can be imported into the
+Terraform state using the `terraform import` command or the [`import`](https://developer.hashicorp.com/terraform/language/import#syntax) block.
```
tech-writer-rob, line 24:
```suggestion
For each connected cluster, a default scaling policy is created. An existing scaling policy can be imported into the
Terraform state using the `terraform import` command or the [`import`](https://developer.hashicorp.com/terraform/language/import#syntax) block (recommended for Terraform 1.5.0+).
```

templates/resources/workload_scaling_policy.md.tmpl
```diff
@@ -18,12 +18,111 @@ simultaneously or create custom policies with different settings and apply them
...
+   ```
+
+2. Run the `terraform import` command:
+
+   ```shell
+   terraform plan -out=import.plan -var-file=tf.vars -generate-config-out=generated.tf
+   ```
+
+3. Review the `generated.tf` file and ensure that the imported scaling policy is correct. Terraform sets zero values,
+	 e.g., for `look_back_period_seconds`. All such properties can be removed to use the defaults.
```
tech-writer-rob, line 47:
```suggestion
3. Review the `generated.tf` file and ensure the imported scaling policy is correct. Terraform will generate this file by setting values equal to zero for certain configuration parameters. 

For example:

   ```hcl
   cpu {
     look_back_period_seconds = 0
   }
```

You can simplify the configuration by:
   - Removing fields set to `0` to use default values
   - Keeping only the fields you want to configure explicitly

tech-writer-rob: The zero values were a little unclear to me

templates/resources/workload_scaling_policy.md.tmpl
```diff
@@ -18,12 +18,111 @@ simultaneously or create custom policies with different settings and apply them
...
+   terraform plan -out=import.plan -var-file=tf.vars
+   ```
+
+3. Apply the import plan:
+
+   ```shell
+   terraform apply "import.plan"
+   ```
+
+### Import using `terraform import` command
```
tech-writer-rob, line 119:
```suggestion
### Import using the `terraform import` command
```

templates/resources/workload_scaling_policy.md.tmpl
```diff
@@ -38,4 +137,10 @@ $ terraform import castai_workload_scaling_policy.services <cluster_id>/services
...
 $ terraform import 'module.castai-eks-cluster.castai_workload_scaling_policy.this["services"]' <cluster_id>/services
-```
\ No newline at end of file
+```
+
+## Upsert scaling policy
+
+The recommended way is to [import](#importing) the scaling policy and then apply the changes to the policy.
+However, if that’s not possible, you can define the default policy resource yourself. The CAST AI Terraform provider,
+will update the existing policy instead of returning an error.
```
tech-writer-rob, line 146:
```suggestion
## Upsert scaling policy

The recommended way is to [import](#importing) the scaling policy and then apply the changes to the policy.
However, if that’s not possible, you can define the default policy resource yourself. The CAST AI Terraform provider
will update the existing policy instead of returning an error.
```

UOndro (approved):

PR #449, opened 2025-01-31
feat: add adaptive threshold

UOndro:
There was one issue with TF sdk being unable to recognize unset and zero values. Kube team recommended setting value to string. 

- I had to rework how validation works - the denominator in req is float and not pointer, so I would have to map it to 
  -1 or something like that, or validate it during mapping
- I choose to validate it during mapping, too me it seems as much cleaner way how to do it, instead of validating req struct
- Because of that I also had to change tests a little bit

```terraform
resource "castai_workload_scaling_policy" "xyz" {
  name              = "xyz"
  cluster_id        = "68f9f4ce-6910-4e66-a545-b32f93a4638f"
  management_option = "MANAGED"
  cpu {
    apply_threshold_strategy {
      type        = "CUSTOM_ADAPTIVE"
      numerator   = 0.4
      denominator = 0.3
      exponent    = 0.5
    }

    min                      = 0.1
    max                      = 1
  }
  memory {
    function = "MAX"
    overhead = 0.35
    apply_threshold_strategy {
      type        = "DEFAULT_ADAPTIVE"
    }

    limit {
      type       = "MULTIPLIER"
      multiplier = 1.5
    }
    management_option = "READ_ONLY"
  }
  startup {
    period_seconds = 240
  }
  downscaling {
    apply_type = "DEFERRED"
  }
  memory_event {
    apply_type = "IMMEDIATE"
  }
  anti_affinity {
  }
}

```

### DIFF
--- a/castai/resource_workload_scaling_policy.go
+++ b/castai/resource_workload_scaling_policy.go
@@ -24,6 +24,10 @@ const (
 	minApplyThresholdValue          = 0.01
 	maxApplyThresholdValue          = 2.5
 	defaultApplyThresholdPercentage = 0.1
+	minNumeratorValue               = 0.0
+	minDenominatorValue             = 0.
+	maxExponentValue                = 1.
+	minExponentValue                = 0.
 )
 
 const (
@@ -31,11 +35,16 @@ const (
 	FieldLimitStrategyType       = "type"
 	FieldLimitStrategyMultiplier = "multiplier"
 
-	DeprecatedFieldApplyThreshold             = "apply_threshold"
-	FieldApplyThresholdStrategy               = "apply_threshold_strategy"
-	FieldApplyThresholdStrategyType           = "type"
-	FieldApplyThresholdStrategyPercentage     = "percentage"
-	FieldApplyThresholdStrategyPercentageType = "PERCENTAGE"
+	DeprecatedFieldApplyThreshold                  = "apply_threshold"
+	FieldApplyThresholdStrategy                    = "apply_threshold_strategy"
+	FieldApplyThresholdStrategyType                = "type"
+	FieldApplyThresholdStrategyPercentage          = "percentage"
+	FieldApplyThresholdStrategyNumerator           = "numerator"
+	FieldApplyThresholdStrategyDenominator         = "denominator"
+	FieldApplyThresholdStrategyExponent            = "exponent"
+	FieldApplyThresholdStrategyPercentageType      = "PERCENTAGE"
+	FieldApplyThresholdStrategyDefaultAdaptiveType = "DEFAULT_ADAPTIVE"
+	FieldApplyThresholdStrategyCustomAdaptiveType  = "CUSTOM_ADAPTIVE"
 )
 
 var (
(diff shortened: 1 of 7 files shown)


castai/resource_workload_scaling_policy.go
```diff
@@ -291,10 +304,54 @@ func workloadScalingPolicyResourceApplyThresholdStrategySchema() *schema.Resourc
...
+			},
+			FieldApplyThresholdStrategyDenominator: {
+				// Terraform SDK cannot distinguish between unset and 0 value, that's why it has to be string.
+				Type:     schema.TypeString,
+				Optional: true,
+				Description: fmt.Sprintf("If %s is close or equal to 0, the threshold will be much bigger for small values."+
+					"For example when numerator, exponent is 1 and denominator is 0 the threshold for 0.5 req. CPU will be 200%%."+
+					"It must be defined for the %s strategy.", FieldApplyThresholdStrategyDenominator, FieldApplyThresholdStrategyCustomAdaptiveType),
+				ValidateDiagFunc: validateConvertableToFloat(),
+			},
```
aivarasatkocaitiscastai, line 322: Maybe we should have consistency for these three parameters? All strings? 🤔

aivarasatkocaitiscastai: Not that we expect this to be the main flow...

castai/resource_workload_scaling_policy.go
```diff
@@ -5,6 +5,7 @@ import (
 	"fmt"
```
aivarasatkocaitiscastai, line 5: would be nice to see tf usage examples in the PR description 🙏

castai/resource_workload_scaling_policy.go
```diff
@@ -308,11 +365,19 @@ func resourceWorkloadScalingPolicyCreate(ctx context.Context, d *schema.Resource
 	}
 
 	if v, ok := d.GetOk("cpu"); ok {
-		req.RecommendationPolicies.Cpu = toWorkloadScalingPolicies(v.([]interface{})[0].(map[string]interface{}))
+		cpu, err := toWorkloadScalingPolicies("cpu", v.([]interface{})[0].(map[string]interface{}))
```
aivarasatkocaitiscastai, line 368: NIT: can we `interface{}` -> `any` would looks nicer 🙏

aivarasatkocaitiscastai: I see it's old code but is it possible to get panic while `v.([]interface{})[0].(map[string]interface{})` ?

docs/resources/workload_scaling_policy.md
```diff
@@ -111,9 +110,17 @@ Required:
 
 - `type` (String) Defines apply theshold strategy type.
 	- PERCENTAGE - recommendation will be applied when diff of current requests and new recommendation is greater than set value
+    - DEFAULT_ADAPTIVE - will pick larger threshold percentage for small workloads and smaller percentage for large workloads.
+    - CUSTOM_ADAPTIVE - works in same way as DEFAULT_ADAPTIVE, but it allows to tweak parameters of adaptive threshold formula: percentage = numerator/(currentRequest + denominator)^exponent
```
aivarasatkocaitiscastai, line 114: Could also mention that default_adaptive is the recommended way to go and customization is left for advanced use cases

aivarasatkocaitiscastai (approved):

castai/resource_workload_scaling_policy.go
UOndro: I thought about it, but from user perspective it is kinda hidden that it is string, so we would only added more complexity to code. It doesn't have to be wrapped in "":

```
    apply_threshold_strategy {
      type        = "CUSTOM_ADAPTIVE"
      numerator   = 0.4
      denominator = 0.3
      exponent    = 0.5
    }
```

castai/resource_workload_scaling_policy.go
UOndro: Schema validation prevents from panics

PR #452, opened 2025-02-06
NOJIRA: Don't cancel existing build jobs (long running) but queue instead.

Tsonov:
At least this way 2 people doing MRs won't be stepping on each other's toes.


Trojan295 (approved):
Did you check, if two tests can run in parallel?

https://github.com/search?q=repo%3Acastai%2Fterraform-provider-castai%20testAcc&type=code
shows that most resource have a random suffix, so I guess should be fine.

PR #467, opened 2025-03-18
Cast-anywhere Terraform configuration

juliette-cast:
Created an tested this config to deploy a minikube cluster to CASt via TF..it deployed everything the script deploys and i add some more configs..it works fine please review and merge


juliette-cast:
fixing it

PR #471, opened 2025-03-20
feat: aks public ip 

aldor007:


castai/resource_node_configuration.go
```diff
@@ -367,6 +370,61 @@ func resourceNodeConfiguration() *schema.Resource {
...
+										ValidateFunc: func(v interface{}, k string) (ws []string, errors []error) {
+											allowedKeys := []string{
+												"FirstPartyUsage",
+												"NetworkDomain",
+												"RoutingPreference",
+											}
+
+											tags := v.(map[string]interface{})
+											for key := range tags {
+												if !lo.Contains(allowedKeys, key) {
```
Tsonov, line 413: Any chance we might need to be case insensitive here?

castai/resource_node_configuration.go
```diff
@@ -367,6 +370,61 @@ func resourceNodeConfiguration() *schema.Resource {
 								},
 							},
 						},
+						FieldNodeConfigurationAKSNetworkSecurityGroup: {
+							Type:        schema.TypeString,
+							Optional:    true,
+							Description: "Network security group to be used for provisioned nodes, if not provided default security group from `castpool` will be used",
```
Tsonov, line 376: Is it possible to have 1 node with multiple NSGs in Azure?

castai/resource_node_configuration.go
```diff
@@ -1197,9 +1298,47 @@ func flattenAKSConfig(config *sdk.NodeconfigV1AKSConfig) []map[string]interface{
...
+
+	m := map[string]interface{}{}
+	if sdkPublicIp.IpPrefix != nil {
+		m["public_ip_prefix"] = *sdkPublicIp.IpPrefix
+	}
+
+	if sdkPublicIp.Tags != nil {
+		tags := make(map[string]interface{})
+		for _, tag := range *sdkPublicIp.Tags {
+			tags[*tag.IpTagType] = *tag.TagValue
```
Tsonov, line 1329: Do we need `lo.FromPtr` or nil check to ignore weird data?

castai/resource_node_configuration.go
aldor007: VMSS API allows single network security group 
image
ATM I don't see need for array here.

castai/resource_node_configuration.go
aldor007: Will add

castai/resource_node_configuration.go
aldor007: I've tested it once again and it allows lower-case values. But if you don't mind I would leave it as is - I would need to update validation on API level.

Tsonov (approved):

PR #472, opened 2025-03-31
WIRE-1038: Add scopes field for role bindings

oanatmaria:


castai/resource_role_bindings.go
```diff
@@ -343,6 +373,31 @@ func assignRoleBindingData(roleBinding *sdk.CastaiRbacV1beta1RoleBinding, data *
...
+						FieldRoleBindingsScopeResourceID: scope.Cluster.Id,
+					},
+				)
+			}
+		}
+	}
+	err := data.Set(FieldRoleBindingsScopes, scopes)
+	if err != nil {
+		return fmt.Errorf("parsing scopes: %w", err)
+	}
```
radekska, line 399: nit
```suggestion
	if err := data.Set(FieldRoleBindingsScopes, scopes); err != nil {
		return fmt.Errorf("parsing scopes: %w", err)
	}
```

castai/resource_role_bindings_test.go
```diff
@@ -230,6 +230,7 @@ role_id = `+roleID+`
 scope.# = 1
 scope.0.kind = organization
 scope.0.resource_id = `+organizationID+`
+scopes.# = 0
```
radekska, line 233: could you add tests with some mocked data?

docs/resources/role_bindings.md
radekska, line 1: btw. could you fix / add an example for `group` as it's missing for some reason in TF docs: https://registry.terraform.io/providers/castai/castai/latest/docs/resources/organization_group 🙏 similar to https://registry.terraform.io/providers/castai/castai/latest/docs/resources/role_bindings

radekska (approved):

PR #476, opened 2025-03-31
Adding cast anywhere TF config

juliette-cast:
This config installs all the components that the cast anywhere script installs via the UI. tested and it works with no issues

### DIFF
--- a/examples/castai-anywhere-terraform/main.tf
+++ b/examples/castai-anywhere-terraform/main.tf
@@ -1,0 +1,163 @@
+# Start Minikube using the Docker driver
+resource "null_resource" "start_minikube" {
+  provisioner "local-exec" {
+    command = "minikube start --driver=docker"
+  }
+}
+
+# Wait for Minikube to be ready
+data "external" "wait_for_minikube" {
+  program    = ["bash", "-c", "while ! kubectl get nodes >/dev/null 2>&1; do sleep 5; done; echo '{}'"]
+  depends_on = [null_resource.start_minikube]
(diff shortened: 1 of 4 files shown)


examples/castai-anywhere-terraform/main.tf
```diff
@@ -0,0 +1,163 @@
...
+    value = var.cluster_name
+  }
+  set {
+    name  = "provider"
+    value = "anywhere"
+  }
+}
+
+# Install CAST AI Cluster Controller
+resource "helm_release" "castai_cluster_controller" {
```
Tsonov, line 68: You probably want to pass `autoscaling.enabled=false` as value as well 🤔  Since anywhere does not support  node autoscaling

examples/castai-anywhere-terraform/variables.tf
```diff
@@ -0,0 +1,32 @@
...
+variable "cluster_id" {
+  description = "Identifier for your CAST AI cluster"
+  type        = string
+  default     = "" # add the cluster ID you copied from the UI from the 1st apply
+}
+
+variable "organization_id" {
+  description = "Your CAST AI Organization ID"
+  type        = string
+  default     = "" # add your Org ID ..this is required for pod mutator deployment
```
Tsonov, line 25: If this must be provided, omit the default to force user to pass a value more clearly

Tsonov (commented):
Minor comments to resolve

Tsonov (approved):

PR #477, opened 2025-04-10
fix: missing node template constraints

primoz-hrovat:
Added missing node template constraints.

Additionally, updated generated API client and handled pointer dereference panic on role bindings.


primoz-hrovat:
Depends on https://github.com/castai/terraform-provider-castai/pull/472 to be merged first due to API changes.

PR #482, opened 2025-04-15
Update README.md

juliette-cast:
Woop has 2 replicas and cannot be placed on same node so starting minikube with 2 or more nodes is the solution


juliette-cast:
@Tsonov i just updated readme to add an instruction

PR #507, opened 2025-06-16
REP-1293: Allocation group resource

roman-castai:
Example usage from the current schema: 
```
resource "castai_allocation_group" "ag_example" {
  name = "ag_example"
  cluster_ids = [
    "1a58d6b4-bc0e-4417-b9c7-31d15c313f3f",
    "d204b988-5db5-472e-a258-bf763a0f4a93"
  ]

  namespaces = [
    "namespace-a",
    "namespace-b"
  ]

  labels = {
    environment              = "production",
    team                     = "my-team",
    "app.kubernetes.io/name" = "app-name"
  }

  labels_operator = "AND"
}
```


castai/resource_allocation_group.go
```diff
@@ -0,0 +1,287 @@
...
+		DeleteContext: resourceAllocationGroupDelete,
+		Importer: &schema.ResourceImporter{
+			StateContext: allocationGroupImporter,
+		},
+		Schema: map[string]*schema.Schema{
+			"name": {
+				Type:             schema.TypeString,
+				Required:         true,
+				Description:      "Allocation group name",
+				ValidateDiagFunc: validation.ToDiagFunc(validation.StringMatch(k8sNameRegex, "name must adhere to the format guidelines of Kubernetes labels/annotations")),
```
zolinas, line 31: CastAI console is not so strict regarding allocation group names. For example "Team Reporting" is a valid allocation group name which would not pass this validation.

castai/resource_allocation_group.go
```diff
@@ -0,0 +1,287 @@
...
+		return diag.FromErr(err)
+	}
+
+	allocationGroups := *response.JSON200.Items
+
+	allocationGroupName := d.Get("name").(*string)
+	if len(allocationGroups) > 0 {
+		for _, ag := range allocationGroups {
+			if *ag.Name == *allocationGroupName {
+				return diag.Errorf("allocation group already exists")
```
zolinas, line 152: At the moment we do not require Unique names for Allocations Groups in the console. We should either add this check to the API as well, or remove it from here.

roman-castai: I will check to see if there is an organization that already uses same name for different labeled allocation group and therefore the change would affect someone, and if not I would add this check to the API.

If we don't enforce this it would be possible to have a terraform configuration with the completely equal two entries, e.g.:
```
resource "cast_ai_allocation_group" "my_allocation_group" {
  clusterIds = ["uuid1", "uuid2"]
  namespaces = ["namespace1"]
}

resource "cast_ai_allocation_group" "my_allocation_group" {
  clusterIds = ["uuid1", "uuid2"]
  namespaces = ["namespace1"]
}
```
We could probably check the whole allocation group definition instead of just the name and would guard from most of the unintended use cases like:

* Given that the id would be implicit it wouldn't be noticeable in the tf config that those are really two different allocation groups.

* If migration of previously created allocation groups is not done through importing but just by copying everything in the tf files, the client would recreate all their allocation groups which I would say we should guard for.

But I would argue checking the name uniqueness would make sense from the start as I presume it would be confusing to have two allocation groups with the same name showing same or even different statistics.

tl;dr: I will first check if anyone uses non-unique allocation group names, if not I would enforce it.

castai/resource_allocation_group.go
```diff
@@ -0,0 +1,287 @@
...
+		},
+		Schema: map[string]*schema.Schema{
+			"name": {
+				Type:        schema.TypeString,
+				Required:    true,
+				Description: "Allocation group name",
+			},
+			"cluster_ids": {
+				Type:        schema.TypeList,
+				Required:    true,
```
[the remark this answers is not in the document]
roman-castai: Got it, wasn't aware clients can create AGs through the API, will update it like that then.

castai/resource_allocation_group.go
```diff
@@ -0,0 +1,303 @@
...
+func resourceAllocationGroup() *schema.Resource {
+	return &schema.Resource{
+		CreateContext: resourceAllocationGroupCreate,
+		ReadContext:   resourceAllocationGroupRead,
+		UpdateContext: resourceAllocationGroupUpdate,
+		DeleteContext: resourceAllocationGroupDelete,
+		CustomizeDiff: resourceAllocationGroupDiff,
+		Description:   "Manage allocation group. Allocation group [reference](https://docs.cast.ai/docs/allocation-groups)",
+		Importer: &schema.ResourceImporter{
+			StateContext: allocationGroupImporter,
```
furkhat, line 27: You seem to just use ID to get the resource. Why custom importer? I think it would work with this
```suggestion
			StateContext: schema.ImportStatePassthroughContext,
```

castai/resource_allocation_group.go
```diff
@@ -0,0 +1,303 @@
...
+	}
+	if err := d.Set("namespaces", ag.Filter.Namespaces); err != nil {
+		return diag.FromErr(fmt.Errorf("setting namespaces: %w", err))
+	}
+	if err := d.Set("labels", fromLabels(*ag.Filter.Labels)); err != nil {
+		return diag.FromErr(fmt.Errorf("setting labels: %w", err))
+	}
+	if err := d.Set("labels_operator", ag.Filter.LabelsOperator); err != nil {
+		return diag.FromErr(fmt.Errorf("setting labels_operator: %w", err))
+	}
```
furkhat, line 145: Please check examples of `mapstructure` annotation. It should do this automatically

PR #519, opened 2025-06-26
Live AWS with native VPC CNI example

apasyniuk:


examples/eks/eks_cluster_live_migration/castai.tf
```diff
@@ -0,0 +1,15 @@
...
+  cluster_name     = var.cluster_name
+  castai_api_token = var.castai_api_token
+  cluster_region   = var.cluster_region
+  vpc_id           = module.vpc.vpc_id
+  security_groups = [
+    module.eks.cluster_security_group_id,
+    module.eks.node_security_group_id,
+    aws_security_group.additional.id,
+  ]
+  subnets = module.vpc.private_subnets
```
aldor007, line 14: depend_on = [ module.eks[

apasyniuk: Unfortuantelly this doesnt work with module.eks I get circural dependencies. with only modle.vpc I get some cryptic error with "for each` something something. So for now it is better that wat was before.

PR #522, opened 2025-06-26
Updating the provider blocks

juliette-cast:
Updating becasue helm expects the kubernetes config to be passed as a map, not a nested block


juliette-cast:
hey @furkhat i fixed the 1st check and this 2nd one failing validation is about credentials and i even tried fixing but same error..please review and advice how to fix it and re-run the checks thank you

juliette-cast:
Hey @furkhat pleasr review this when you can unless the module has been fixed thank you

PR #527, opened 2025-07-02
WOOP-867 Add rollout behavior to workload optimization scaling policy

zrudzionis:
### Description
Add rollout behavior to workload optimization scaling policy

```
  rollout_behavior {
    type = "NO_DISRUPTION"
  }
```


castai/resource_workload_scaling_policy.go
```diff
@@ -268,6 +271,23 @@ It can be either:
...
+						FieldType: {
+							Type:     schema.TypeString,
+							Required: true,
+							Description: `Defines the rollout type to be used when applying recommendations.
+	- NO_DISRUPTION - pods are restarted without causing service disruption.`,
+							ValidateDiagFunc: validation.ToDiagFunc(validation.StringInSlice([]string{FieldRolloutBehaviorNoDisruptionType}, false)),
+						},
+					},
+				},
+			},
```
aivarasatkocaitiscastai, line 290: Maybe worth adding notes about this being supported for single replica, like we have in UI

castai/resource_workload_scaling_policy.go
```diff
@@ -36,6 +36,9 @@ const (
 	FieldLimitStrategyType                         = "type"
 	FieldLimitStrategyMultiplier                   = "multiplier"
 	FieldConfidence                                = "confidence"
+	FieldRolloutBehavior                           = "rollout_behavior"
+	FieldType                                      = "type"
```
aivarasatkocaitiscastai, line 40: Other `type` values have more descriptive names, is this generic on purpose?

aivarasatkocaitiscastai (approved):

castai/resource_workload_scaling_policy.go
zrudzionis: Will add link to the docs with prerequisites: https://docs.cast.ai/docs/workload-autoscaling-configuration#prerequisites.

castai/resource_workload_scaling_policy.go
zrudzionis: Will rename it to `FieldRolloutBehaviorType`. Probably could also use `FieldType ` as it is unlikely that we will rename fields as this would be a breaking change.

Makefile
```diff
@@ -66,7 +66,7 @@ test:
 .PHONY: testacc
 testacc:
 	@echo "==> Running acceptance tests"
-	TF_ACC=1 go test ./castai/... '-run=^TestAcc' -v -timeout 50m
+	TF_LOG=INFO TF_ACC=1 go test -count=1 ./castai/... '-run=^TestAccResourceWorkloadScalingPolicy' -v -timeout 50m
```
zrudzionis, line 69: Will revert this before merging.

PR #539, opened 2025-07-24
fix: allow deleting policies with assignment rules

mKeRix:
Today, workload scaling policies with assignment rules managed by Terraform cannot be deleted like any other resource. Since these rules assign workloads to the policies, deleting them fails with the error `cannot delete the policy as workloads are assigned to it`. This does not make a lot of sense though: the assignment rules are managed in the same Terraform resource, so a delete operation intentionally also wants to delete the assignment rules.

This change fixes this behavior by making deletion a two step process, where assignment rules are first taken away and after the actual deletion occurs.

This is a much cleaner and more expected behavior vs having to manually null out all assignment rules before deletion.

This can be tested using a Terraform config like this (running apply and then destroy):

```terraform
terraform {
  required_providers {
    castai = {
      source  = "castai/castai"
      version = ">= 7.38.0, < 8.0.0"
    }
  }
}

locals {
  castai_cluster_id = "<insert-id>"
}

resource "castai_workload_scaling_policy" "balanced" {
  name              = "custom-balanced"
  cluster_id        = local.castai_cluster_id
  management_option = "MANAGED"

  cpu {
    overhead                 = 0
    args                     = ["0.95"]
    min                      = 0.01
    look_back_period_seconds = 86400

    limit {
      type = "NO_LIMIT"
    }

    apply_threshold_strategy {
      type = "DEFAULT_ADAPTIVE"
    }
  }

  memory {
    function                 = "MAX"
    overhead                 = 0.1
    min                      = 10
    look_back_period_seconds = 86400

    apply_threshold_strategy {
      type = "DEFAULT_ADAPTIVE"
    }
  }

  downscaling {
    apply_type = "IMMEDIATE"
  }

  startup {
    period_seconds = 5 * 60
  }

  anti_affinity {
  }

  confidence {
    threshold = 0.9
  }

  assignment_rules {
    rules {
      workload {
        gvk = ["Deployment", "Rollout", "DaemonSet"]
      }
    }
  }
}

resource "castai_workload_scaling_policy" "stability" {
  name              = "custom-stability"
  cluster_id        = local.castai_cluster_id
  management_option = "MANAGED"

  cpu {
    overhead                 = 0.1
    args                     = ["0.99"]
    min                      = 0.01
    look_back_period_seconds = 86400

    limit {
      type = "NO_LIMIT"
    }

    apply_threshold_strategy {
      type = "DEFAULT_ADAPTIVE"
    }
  }

  memory {
    function                 = "MAX"
    overhead                 = 0.1
    min                      = 10
    look_back_period_seconds = 86400

    apply_threshold_strategy {
      type = "DEFAULT_ADAPTIVE"
    }
  }

  downscaling {
    apply_type = "IMMEDIATE"
  }

  anti_affinity {
  }

  confidence {
    threshold = 0.9
  }

  assignment_rules {
    rules {
      namespace {
        names = ["kube-system"]
      }
    }
  }
}
```

### DIFF
--- a/castai/resource_workload_scaling_policy.go
+++ b/castai/resource_workload_scaling_policy.go
@@ -2,6 +2,7 @@ package castai
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"net/http"
 	"regexp"
@@ -9,6 +10,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/retry"
 	"github.com/samber/lo"
 
 	"github.com/google/uuid"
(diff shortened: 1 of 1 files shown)


mszostok:
Hi, thanks for the contribution. 
We completely agree that this is a real issue and needs to be addressed. However, we are evaluating alternative options for handling this case.

I will get back to you as soon as possible on whether we will proceed with this approach or address it globally on our side.

mszostok:
We’ll be removing the validation on our backend side, so I'm going to close this pull-request. Again, thanks for the contribution and valuable feedback 🙏

PR #546, opened 2025-07-31
NOJIRA: Fix resourceCastaiSSOConnectionRead after proto change

oanatmaria:


radekska (commented):
Doesn't it break the UI as well?

PR #568, opened 2025-09-23
Update README.md

juliette-cast:
Just adding that prediction is only support on AWS for now so customers are aware


dvoros-castai (approved):

pijusn-cast:
Rebased this PR to latest `master` to fix the build error in the checks. Note that I see this repository has broken acceptance tests for a while now. That means build will fail on the acceptance tests but that does not prevent the PR to get merged.

PR #569, opened 2025-09-24
chore: make generate-sdk

pijusn-cast:
Running `make generate-sdk` on the repository because it causes PRs to fail ([example](https://github.com/castai/terraform-provider-castai/pull/568)).


laimonasr (approved):

pijusn-cast:
acceptance tests are broken but that seems to be the case for a while (I see `master` contains PRs merged with broken acceptance tests with the same error already. I guess since it's optional, nobody really cares.

PR #662, opened 2026-03-12
PRICE-553: add terraform support for pod mutations

pawelpeksa:
PRICE-553: add terraform support for pod mutations


castai/resource_pod_mutation.go
```diff
@@ -0,0 +1,1316 @@
...
+			MaxItems: 1,
+			Elem: &schema.Resource{
+				Schema: configSchema,
+			},
+		},
+	}
+}
+
+func podMutationStateImporter(ctx context.Context, d *schema.ResourceData, meta any) ([]*schema.ResourceData, error) {
+	organizationID, id := parseImportID(d)
```
michal-kopczynski: `parseImportID` is defined in resource_hibernation_schedule.go. Maybe it should be moved to some shared file?

castai/resource_pod_mutation.go
```diff
@@ -0,0 +1,1316 @@
...
+							if exprs, ok := prefNodeMap[FieldPodMutationMatchExpressions]; ok {
+								exprList := exprs.([]interface{})
+								reqs := make([]patching_engine.NodeSelectorRequirement, 0, len(exprList))
+								for _, e := range exprList {
+									if e == nil {
+										continue
+									}
+									em := e.(map[string]interface{})
+									key := em[FieldPodMutationTolerationKey].(string)
+									operator := em[FieldPodMutationTolerationOperator].(string)
```
michal-kopczynski: should `NodeSelectorRequirement` have it own specific key and operator constants, and not re use the one from toleration?

castai/resource_pod_mutation.go
```diff
@@ -0,0 +1,1304 @@
...
+			ForceNew:         true,
+			Description:      "Name of the pod mutation.",
+			ValidateDiagFunc: validation.ToDiagFunc(validation.StringIsNotWhiteSpace),
+		},
+		FieldPodMutationEnabled: {
+			Type:        schema.TypeBool,
+			Required:    true,
+			Description: "Whether the pod mutation is enabled.",
+		},
+		FieldPodMutationObjectFilterV2: {
```
michal-kopczynski: I can see that  TF is following the API naming conventions rather than the Custom Resource ones. For example, filters in the API (and now also TF) use `label_filter` and `kind`, while Custom Resources use `pod.labels` and `workload.kinds`.

I get why — TF uses the API under the hood, so aligning with it makes sense. But on the other hand, the Custom Resource naming is newer and arguably improved, so it might also make sense to use that as the standard in TF instead. My guess is more customers use CRs directly than the API, so they're probably more familiar with that convention making transition easier 🤔

Also, does the ...V2 suffix make sense if there's no V1 in this interface?

But feel free to challenge it. On the other hand, having three interfaces with incompatible naming is also not ideal, from maintenance POV. 🤔

castai/resource_pod_mutation.go
```diff
@@ -0,0 +1,1349 @@
...
+				if err != nil {
+					return nil, fmt.Errorf("flattening tolerations: %w", err)
+				}
+
+				configMap[FieldPodMutationTolerations] = flatTolerations
+			}
+			if g.Config.Affinity != nil {
+				configMap[FieldPodMutationAffinity] = flattenAffinity(g.Config.Affinity)
+			}
+			if g.Config.SpotType != nil && *g.Config.SpotType != "SPOT_TYPE_UNSPECIFIED" {
```
michal-kopczynski: Please use typed constant from the sdk.

castai/resource_pod_mutation_test.go
```diff
@@ -0,0 +1,782 @@
...
+			SpotDistributionPercentage: lo.ToPtr(int32(80)),
+			Source:                     lo.ToPtr(patching_engine.API),
+			ObjectFilterV2: &patching_engine.ObjectFilterV2{
+				Namespaces: &[]patching_engine.ObjectFilterV2Matcher{
+					{Type: lo.ToPtr(patching_engine.EXACT), Value: lo.ToPtr("default")},
+				},
+				Kinds: &[]patching_engine.ObjectFilterV2Matcher{
+					{Type: lo.ToPtr(patching_engine.EXACT), Value: lo.ToPtr("Deployment")},
+				},
+			},
```
michal-kopczynski, line 99: is there also some test verifying other fields of object filter, especially labels_filter and also exclusions?

michal-kopczynski (approved):

castai/resource_pod_mutation_test.go
pawelpeksa: added

castai/resource_pod_mutation.go
pawelpeksa: Updated to have this distinction, I also introduced couple more changes, got rid of "restartMatchingWorkloads" since it's not something we support yet and rename object filter v2 to filter v2.

Regarding v1 and v2. Even if v1 is not exposed, I lean toward leaving `v2` suffix, it doesn't change anything for the user I think, also, users are technical so it's not increased cognitive load because need of versioning is obvious for them I believe. 

I didn't expose patches v2 because it's not supported by API yet (you mentioned it in different comment). Maybe it would make sense to do that but since it's addition, I think it's going to be easy to implement on top of this. We'll see if we get any questions on that as well. 

I also considered leaving only patches, as CRD does, but users migrating to this will be user migration from UI so they likely don't want to rework all their mutations to patches I think

castai/resource_pod_mutation.go
```diff
@@ -0,0 +1,1501 @@
...
+			Optional:    true,
+			Description: "Distribution groups for percentage-based pod distribution.",
+			Elem: &schema.Resource{
+				Schema: distributionGroupSchema(),
+			},
+		},
+		FieldPodMutationSource: {
+			Type:        schema.TypeString,
+			Computed:    true,
+			Description: "Source of the pod mutation (API or CUSTOM_RESOURCE).",
```
michal-kopczynski, line 455: There could be an indication in the description that this field is read only.

examples/resources/castai_pod_mutation/resource.tf
```diff
@@ -0,0 +1,119 @@
+# Basic pod mutation that adds spot scheduling to workloads in the "default" namespace.
+resource "castai_pod_mutation" "spot_scheduling" {
+  cluster_id = castai_eks_cluster.example.id
+  name       = "spot-scheduling"
+  enabled    = true
+
+  filter_v2 {
+    workload {
```
michal-kopczynski, line 8: Nit: not saying examples have to show every possible combination, but would be nice to enhance them a bit e.g. show workload names filler and at least one exclude filter in some example , at the very least.

michal-kopczynski (approved):

castai/resource_pod_mutation.go
pawelpeksa: it's `computed`. Doesn't it imply that it's read only? I see it gets also generated like this so I would rather not add it to the description on top of that:
image

examples/resources/castai_pod_mutation/resource.tf
pawelpeksa: definitely, added examples

castai/resource_pod_mutation.go
michal-kopczynski: Fair enough. I was mislead by the fact that OrganizationID is Computed: true as well, yet it's not read only.

PR #679, opened 2026-04-02
LLM-1102: AI Optimizer Hosted Model Resources

krzysztofreczek:
## Summary

  This PR adds Terraform support for deploying and managing self-hosted models through CAST AI's AI Optimizer. It
  introduces three new resources and a generated SDK client.

  ### New Resources

  #### `castai_ai_optimizer_hosted_model`

  The core resource for deploying a model onto a CAST AI-managed cluster. Key capabilities:

  - Deploys a model as a Kubernetes service on a specified cluster and port
  - References a model specs definition (predefined or custom)
  - **vLLM configuration** -- HuggingFace token injection via a Kubernetes secret or inline token
  - **Horizontal autoscaling**
  - **Hibernation**
  - **Fallback**
  - Node template selection and region placement
  - Full CRUD + import support

  #### `castai_ai_optimizer_model_specs`

  Defines the model specification (what model to serve). Supports two registry types:

  - **HuggingFace** -- reference a public HuggingFace model by name (e.g., `meta-llama/Llama-3.1-8B-Instruct`)
  - **Private registry** -- reference a model stored in a private S3-based model registry

  Includes metadata: model type (e.g., `chat`), description, and whether the model is routable.

  #### `castai_ai_optimizer_model_registry`

  Registers a private S3 bucket as a model store. Configured with bucket name, AWS region, optional path prefix, and
  credentials.

  ### Supporting Changes

  - Allow defining an `organization_id` in case the provided CastAI token has access to more than one
  - New `ai_optimizer` SDK package with generated OpenAPI client, models, and mocks
  - Full EKS example (`eks_cluster_existing_with_ai_optimizer`) demonstrating end-to-end setup
  - Provider registration in both legacy and framework-based provider entrypoints
  - Documentation for all three resources


castai/resource_ai_optimizer_hosted_model.go
```diff
@@ -0,0 +1,707 @@
...
+	if err != nil {
+		return diag.FromErr(fmt.Errorf("fetching organization ID: %w", err))
+	}
+
+	clusterID := d.Get(fieldAIHostedModelClusterID).(string)
+
+	tflog.Debug(ctx, "Deleting AI hosted model", map[string]any{"id": d.Id()})
+
+	resp, err := client.HostedModelsAPIDeleteHostedModelWithResponse(ctx, orgID, clusterID, d.Id())
+	if err := sdk.CheckOKResponse(resp, err); err != nil {
```
pmateusz, line 445: All three delete functions treat a 404 response from the API as a hard error. If a resource is already deleted out-of-band (e.g., someone deleted it via the UI or another tool), running terraform destroy will fail instead of succeeding gracefully.

In Terraform, delete should be idempotent. If the resource is already gone, that's the desired state - the delete should succeed. Other resources in this codebase (e.g., resource_service_account_key.go:126) follow this pattern by checking for 404 before treating it as an error.

How to fix: In each delete function, after the API call, check if the status code is 404 and return nil (success) in that case.

```go
  resp, err := client.HostedModelsAPIDeleteHostedModelWithResponse(ctx, orgID, clusterID, d.Id())
  if resp != nil && resp.StatusCode() == http.StatusNotFound {
      return nil
  }
  if err := sdk.CheckOKResponse(resp, err); err != nil {
      return diag.FromErr(fmt.Errorf("deleting hosted model: %w", err))
  }
```

The issue flagged here is a codebase-wide pattern, so the issue is rather low. Most delete functions in this codebase also don't handle 404.

castai/utils.go
```diff
@@ -142,19 +142,24 @@ func normalizeJSON(bytes []byte) ([]byte, error) {
 }
 
 func getDefaultOrganizationId(ctx context.Context, meta any) (string, error) {
```
pmateusz, line 144: solid improvement!

castai/resource_ai_optimizer_hosted_model.go
```diff
@@ -0,0 +1,707 @@
...
+
+func resourceAIHostedModelRead(ctx context.Context, d *schema.ResourceData, meta any) diag.Diagnostics {
+	client := meta.(*ProviderConfig).aiOptimizerClient
+
+	orgID, err := getDefaultOrganizationId(ctx, meta)
+	if err != nil {
+		return diag.FromErr(fmt.Errorf("fetching organization ID: %w", err))
+	}
+
+	clusterID := d.Get(fieldAIHostedModelClusterID).(string)
```
pmateusz, line 316: nit: only happens when terraform state is manually corrupted.

if clusterID is an empty string, the API call becomes something like `GET /v1/organizations/org-1/clusters//hosted-models`

It will result in 404 which is handled, but the error message is not accurate.

```go
clusterID := d.Get(fieldAIHostedModelClusterID).(string)
if clusterID == "" {
      return diag.Errorf("cluster_id is required but was empty in state")
}
```

Same applies to resourceAIHostedModelUpdate (line 395) and resourceAIHostedModelDelete (line 428).

castai/resource_ai_optimizer_hosted_model.go
```diff
@@ -0,0 +1,707 @@
...
+	return setAIHostedModelData(d, model)
+}
+
+func findHostedModelByID(ctx context.Context, client ai_optimizer.ClientWithResponsesInterface, orgID, clusterID, modelID string) (*ai_optimizer.HostedModel, error) {
+	var cursor *string
+	for {
+		params := &ai_optimizer.HostedModelsAPIListHostedModelsParams{
+			PageCursor: cursor,
+		}
+		resp, err := client.HostedModelsAPIListHostedModelsWithResponse(ctx, orgID, clusterID, params)
```
pmateusz, line 350: correct approach given our API limitations, no API for getting a model by ID

examples/eks/eks_cluster_with_ai_optimizer/outputs.tf
```diff
@@ -0,0 +1,29 @@
...
+  description = "EKS cluster endpoint"
+  value       = module.eks.cluster_endpoint
+}
+
+output "cluster_name" {
+  description = "EKS cluster name"
+  value       = module.eks.cluster_name
+}
+
+output "predicted_model_service" {
```
pmateusz: predefined_model_serivice 🤔  - highlighted by a spell checker

pmateusz (approved):

examples/eks/eks_cluster_with_ai_optimizer/outputs.tf
krzysztofreczek: fixed in: https://github.com/castai/terraform-provider-castai/pull/679/changes/0176789b66a93e95d30652bceef9902d4fe9b48a

castai/resource_ai_optimizer_hosted_model.go
krzysztofreczek: fixed: https://github.com/castai/terraform-provider-castai/pull/679/changes/0176789b66a93e95d30652bceef9902d4fe9b48a

castai/resource_ai_optimizer_hosted_model.go
krzysztofreczek: fixed: https://github.com/castai/terraform-provider-castai/pull/679/changes/0176789b66a93e95d30652bceef9902d4fe9b48a

PR #681, opened 2026-04-08
KUBE-1913: [aks] accelerated networking

ValyaB:
Summary                                                                                                                                                                                                                                                                                                                   
                                                                                                                                                                                                                                                                                                                            
  - Adds accelerated_networking field to the AKS node configuration resource, allowing users to control SR-IOV accelerated networking on node NICs                                                                                                                                                                          
  - Supported values: disabled. If not specified: enable when the VM SKU supports it                                                                         
                                                                                                                                       
                                                                                                                                                                                                                                                                                                                            
  Changes                                                                                                                                                                                                                                                                                                                   
                                                                                                                                                                                                                                                                                                                            
  - New schema field accelerated_networking under aks block (Optional, no provider-side default)                                                                                                                                                                                                                            
  - toAKSAcceleratedNetworkingMode / fromAKSAcceleratedNetworkingMode converters with explicit cases and nil default (consistent with toAKSOSDiskType, toAKSImageFamily)                                                                                                                                                                                                                                                                                                                                                                                                            
  - Unit tests for both flatten and expand paths, including nil/empty cases

### DIFF
--- a/castai/resource_node_configuration.go
+++ b/castai/resource_node_configuration.go
@@ -66,6 +66,8 @@ const (
 	aksEphemeralDiskPlacementCacheDisk    = "cacheDisk"
 	aksEphemeralDiskPlacementResourceDisk = "resourceDisk"
 	aksEphemeralDiskPlacementNVME         = "nvmeDisk"
+	aksAcceleratedNetworkingEnabledIfSupported = "enabled_if_supported"
+	aksAcceleratedNetworkingDisabled           = "disabled"
 )
 
 const (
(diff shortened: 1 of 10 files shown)


castai/resource_node_configuration.go
```diff
@@ -497,6 +499,12 @@ func resourceNodeConfiguration() *schema.Resource {
 							Optional:    true,
 							Description: "Whether to enable encryption at host for provisioned nodes. See https://learn.microsoft.com/en-us/azure/virtual-machines/disk-encryption#encryption-at-host---end-to-end-encryption-for-your-vm-data",
 						},
+						FieldNodeConfigurationAKSAcceleratedNetworking: {
+							Type:             schema.TypeString,
+							Optional:         true,
+							Description:      "Controls SR-IOV accelerated networking on the node NIC. Allowed values: `disabled` (force off regardless of SKU capability). When omitted, the field is not sent to the API and the API default applies.",
```
cirisked, line 505: > When omitted, the field is not sent to the API and the API default applies.

What is API default - `enabled`? Why not to include it in allowed values?

ValyaB: unspecifid - API will apply default AZURE logic: if sku supports EP will enable it

ValyaB: So, specify only when to disable; otherwise, the backend decides whether it is enabled.

cirisked (approved):

PR #697, opened 2026-05-11
CSU-5202: Clarify aggressive mode configuration parameters

oleksandrsan:
aggressive_mode_config takes precedence only when the legacy aggressive_mode is false

---
### What changed
Emits a Terraform warning when the legacy `aggressive_mode` parameter shadows the newer `aggressive_mode_config` block in `castai_rebalancing_schedule`, and clarifies field descriptions to explain precedence.

### Why
Setting both options simultaneously causes silent shadowing and a perpetual diff, because `aggressive_mode=true` overrides the granular config. This surfaces the conflict to the user and steers them toward the recommended `aggressive_mode_config`.

### Key changes
- `castai/resource_rebalancing_schedule.go`
  - Added `warnIfAggressiveModeShadowsConfig` to return a diagnostic warning when both `aggressive_mode=true` and `aggressive_mode_config` are present.
  - Updated `resourceRebalancingScheduleCreate` and `resourceRebalancingScheduleUpdate` to invoke the check and append warnings to the result.
  - Refined schema descriptions for `aggressive_mode` and `aggressive_mode_config` to document precedence.
- `docs/resources/rebalancing_schedule.md`
  - Updated argument descriptions and added an `aggressive_mode_config` example block.
- `examples/resources/castai_rebalancing_schedule/resource.tf`
  - Added an example `aggressive_mode_config` block.

### Impact
During create and update operations, Terraform now surfaces a warning if the legacy `aggressive_mode` flag is set alongside `aggressive_mode_config`, indicating that the config block is ignored and providing remediation steps. This is fully backward compatible.

### DIFF
--- a/castai/resource_rebalancing_schedule.go
+++ b/castai/resource_rebalancing_schedule.go
@@ -282,7 +282,7 @@ func warnIfAggressiveModeShadowsConfig(d *schema.ResourceData) diag.Diagnostics 
 	return diag.Diagnostics{{
 		Severity:      diag.Warning,
 		Summary:       "aggressive_mode_config is ignored while aggressive_mode = true",
-		Detail:        "The legacy `aggressive_mode` overrides `aggressive_mode_config`, so the granular settings will not be applied and Terraform will report a perpetual diff. Set `aggressive_mode = false` to use `aggressive_mode_config`, or remove the `aggressive_mode_config` block.",
+		Detail:        "The legacy `aggressive_mode` overrides `aggressive_mode_config`, so the granular settings will not be applied and Terraform will report a perpetual diff. Set `aggressive_mode=false` to use `aggressive_mode_config`, or remove the `aggressive_mode_config` block.",
 		AttributePath: cty.GetAttrPath("launch_configuration").IndexInt(0).GetAttr("aggressive_mode_config"),
 	}}
 }
--- a/castai/resource_rebalancing_schedule.go
+++ b/castai/resource_rebalancing_schedule.go
@@ -8,6 +8,7 @@ import (
 	"time"
 
 	"github.com/google/uuid"
+	"github.com/hashicorp/go-cty/cty"
 	"github.com/hashicorp/terraform-plugin-log/tflog"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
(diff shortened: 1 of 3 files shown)


castai/resource_rebalancing_schedule.go
```diff
@@ -120,7 +120,7 @@ func resourceRebalancingSchedule() *schema.Resource {
 						"aggressive_mode": {
 							Type:        schema.TypeBool,
 							Optional:    true,
-							Description: "When enabled rebalancing will also consider problematic pods (pods without controller, job pods, pods with removal-disabled annotation) as not-problematic.",
+							Description: "When enabled, rebalancing considers all problematic pods (pods without controller, job pods, pods with removal-disabled annotation) as not-problematic — equivalent to enabling every option in `aggressive_mode_config`.",
```
vladklokun: issue: this isn't true

> equivalent to enabling every option in `aggressive_mode_config`

Local persistent volumes are not included in the legacy aggressive mode definition and would not be set to `true` in the new aggressive mode config during normalization.

I’m fine with the other part of this edit.

castai/resource_rebalancing_schedule.go
```diff
@@ -150,7 +150,7 @@ func resourceRebalancingSchedule() *schema.Resource {
 									},
 								},
 							},
-							Description: "Advanced configuration for aggressive rebalancing mode.",
+							Description: "Advanced configuration for aggressive rebalancing mode. Only takes effect when `aggressive_mode = false` (or unset); while the `aggressive_mode` is `true`, these values are ignored by the backend.",
```
vladklokun: suggestion: instruct customers to migrate to the new option.

We should add a call to action: ask customers to turn off the legacy option.

```suggestion
							Description: "Advanced configuration for the aggressive rebalancing mode. This is the recommended way to configure aggressive rebalancing. Please keep the `aggressive_mode` parameter unset or set it `aggressive_mode=false` before using this config option. When the legacy `aggressive_mode` is set to `true`, it takes precedence over this option.",
```

vladklokun (changes requested):

KarolisL:
Would it be possible to emit a warning when the user sets both  `aggressive_mode_config` and `aggressive_mode=true`?

oleksandrsan:
> Would it be possible to emit a warning when the user sets both `aggressive_mode_config` and `aggressive_mode=true`?

Yeah, but only on apply with SDK v2, cause it does not support diagnostics interface in resource validation. We would need to switch to the newer plugin framework to use it.
I've added warning to apply phase, currently it looks like this:
```
castai_rebalancing_schedule.rebalance["postscale"]: Modifying... [id=6881fdb1-8396-45d0-ae1b-f78468ffa1d0]
castai_rebalancing_schedule.rebalance["postscale"]: Modifications complete after 0s [id=6881fdb1-8396-45d0-ae1b-f78468ffa1d0]
╷
│ Warning: aggressive_mode_config is ignored while aggressive_mode = true
│
│   with castai_rebalancing_schedule.rebalance["postscale"],
│   on main.tf line 81, in resource "castai_rebalancing_schedule" "rebalance":
│   81:     aggressive_mode_config {
│
│ The legacy `aggressive_mode = true` overrides `aggressive_mode_config`, so the granular settings will not be applied and Terraform will report a perpetual
│ diff. Set `aggressive_mode = false` to use `aggressive_mode_config`, or remove the `aggressive_mode_config` block.
╵

Apply complete! Resources: 0 added, 1 changed, 0 destroyed.
```

vladklokun (approved):

PR #700, opened 2026-05-18
feast: add support for Edge Configurations in CAST AI terraform provider

andrejatcastai:
---
### What changed
Add Terraform resources and data source for managing CAST AI edge configurations. Introduces the `castai_edge_configuration` resource and data source, along with a `castai_edge_configuration_default` marker resource to designate the default configuration for an edge location.

### Why
Enable users to define and version-control edge deployment settings—such as VM image, boot disk size, CRI socket path, and cloud-init user data—for AWS, GCP, OCI, and custom cloud environments.

### Key changes
- `Makefile`: Added `EdgeConfigurationsAPI` to the OMNI SDK code generation spec.
- `castai/resource_edge_configuration.go`: New resource for full CRUD of edge configurations. Supports cloud-specific blocks (`aws`, `gcp`, `oci`, `custom`) with an `ExactlyOneOf` validator, plus `user_data_base64` and `cri` attributes. Supports import by `organization_id/cluster_id/edge_location_id/{id|name}`. Delete is skipped if the API reports the configuration as the default.
- `castai/resource_edge_configuration_default.go`: New resource that sets an existing edge configuration as the default. Delete is a no-op to avoid unintentionally removing the default designation.
- `castai/data_source_edge_configuration.go`: New data source to retrieve an edge configuration by `name`.
- `castai/provider_framework.go`: Registered `newEdgeConfigurationResource`, `newEdgeConfigurationDefaultResource`, and `newEdgeConfigurationDataSource`.
- `castai/sdk/omni/*.gen.go`: Regenerated SDK to include `EdgeConfigurationsAPI` client methods and request/response types.


castai/data_source_edge_configurations.go
```diff
@@ -0,0 +1,377 @@
...
+
+func newEdgeConfigurationsDataSource() datasource.DataSource {
+	return &edgeConfigurationsDataSource{}
+}
+
+func (d *edgeConfigurationsDataSource) Metadata(_ context.Context, req datasource.MetadataRequest, resp *datasource.MetadataResponse) {
+	resp.TypeName = req.ProviderTypeName + "_edge_configurations"
+}
+
+func (d *edgeConfigurationsDataSource) Schema(_ context.Context, _ datasource.SchemaRequest, resp *datasource.SchemaResponse) {
```
varnastadeus: why do we have data source that return multiple edge configurations while all others are in scope of single resource? whats the use case here? also do we need data source for edge configurations at all?

castai/resource_edge_configuration.go
```diff
@@ -0,0 +1,735 @@
...
+
+	if apiResp.StatusCode() != http.StatusOK {
+		resp.Diagnostics.AddError(
+			"Failed to read edge configuration",
+			fmt.Sprintf("unexpected status code: %d, body: %s", apiResp.StatusCode(), string(apiResp.Body)),
+		)
+		return
+	}
+
+	plan := r.edgeConfigurationToTFModel(apiResp.JSON200, state.OrganizationID, state.ClusterID)
```
varnastadeus: `state`

castai/resource_edge_configuration.go
```diff
@@ -0,0 +1,735 @@
...
+	}
+	if config.Aws != nil {
+		state.AWS = r.toAWSConfigurationModel(config.Aws)
+	}
+	if config.Oci != nil {
+		state.OCI = r.toOCIConfigurationModel(config.Oci)
+	}
+	if config.Custom != nil {
+		state.Custom = r.toCustomConfigurationModel(config.Custom)
+	}
```
varnastadeus: the nil check are redunant as each of the helper methods already do nil check, you can just do directly

```go
{
    Aws: r.toAWSConfigurationModel(config.Aws)
}
```

castai/resource_edge_configuration.go
```diff
@@ -0,0 +1,735 @@
...
+	if plan.AWS != nil {
+		createReq.Aws = r.toAWSConfiguration(plan.AWS)
+	}
+	if plan.OCI != nil {
+		createReq.Oci = r.toOCIConfiguration(plan.OCI)
+	}
+	if plan.Custom != nil {
+		createReq.Custom = r.toCustomConfiguration(plan.Custom)
+	}
```
varnastadeus: same here

castai/resource_edge_configuration.go
```diff
@@ -0,0 +1,735 @@
...
+
+	return createReq
+}
+
+func (r *edgeConfigurationResource) edgeConfigurationUpdateToSDK(plan edgeConfigurationModel) omni.EdgeConfigurationsAPIUpdateEdgeConfigurationJSONRequestBody {
+	updateReq := omni.EdgeConfigurationUpdate{
+		Name: lo.ToPtr(plan.Name.ValueString()),
+	}
+
+	updateReq.Default = lo.ToPtr(false)
```
varnastadeus: lets just omit this from update method, there is no reason to always set false

castai/resource_edge_configuration.go
```diff
@@ -0,0 +1,735 @@
...
+
+	if !plan.UserDataBase64.IsNull() {
+		updateReq.UserDataBase64 = lo.ToPtr(plan.UserDataBase64.ValueString())
+	}
```
varnastadeus: we should be sending this field always, if we omit and user removes it from configuration the old value will be used as we are using patch on the backend

castai/resource_edge_configuration.go
```diff
@@ -0,0 +1,735 @@
...
+	if plan.AWS != nil {
+		updateReq.Aws = r.toAWSConfiguration(plan.AWS)
+	}
+	if plan.OCI != nil {
+		updateReq.Oci = r.toOCIConfiguration(plan.OCI)
+	}
+	if plan.Custom != nil {
+		updateReq.Custom = r.toCustomConfiguration(plan.Custom)
+	}
```
varnastadeus: redundant nil checks

castai/resource_edge_configuration.go
```diff
@@ -0,0 +1,735 @@
...
+		createReq.Oci = r.toOCIConfiguration(plan.OCI)
+	}
+	if plan.Custom != nil {
+		createReq.Custom = r.toCustomConfiguration(plan.Custom)
+	}
+
+	return createReq
+}
+
+func (r *edgeConfigurationResource) edgeConfigurationUpdateToSDK(plan edgeConfigurationModel) omni.EdgeConfigurationsAPIUpdateEdgeConfigurationJSONRequestBody {
```
varnastadeus: please check update behaviour for aws/gcp/oci inner objects as with patch we should be able to "reset" fields if they are no longer in the tf config

castai/resource_edge_configuration.go
```diff
@@ -0,0 +1,735 @@
...
+						Description: "Instance/VM labels",
+					},
+					"image_id": schema.StringAttribute{
+						Optional:    true,
+						Description: "Exact image ID (for example 'projects/castai-public-339919/global/images/ubuntu-2404-lts-amd64-cuda') or image family (for example `projects/ubuntu-os-cloud/global/images/family/ubuntu-2404-lts-amd64`) to be used for edge creation",
+					},
+					"boot_disk_size_gib": schema.Int64Attribute{
+						Optional:    true,
+						Description: "Boot disk size in GiB",
+					},
```
varnastadeus, line 161: does it make sense to have this optional? lets say user sets in 500 gb on create and then removes the value from tf config, what would happen?

castai/resource_edge_configuration.go
```diff
@@ -0,0 +1,735 @@
...
+}
+
+func (m *ociConfigurationModel) Equal(other *ociConfigurationModel) bool {
+	if m == nil || other == nil {
+		return m == other
+	}
+	return m.ImageID.Equal(other.ImageID) &&
+		m.BootDiskSizeGiB.Equal(other.BootDiskSizeGiB) &&
+		m.Tags.Equal(other.Tags)
+}
```
varnastadeus: unused

castai/resource_edge_configuration_default.go
```diff
@@ -0,0 +1,321 @@
...
+		Description: "Manages the default CAST AI Edge Configuration for an edge location. " +
+			"This resource sets an existing edge configuration as the default for the specified edge location.",
+		Attributes: map[string]schema.Attribute{
+			"id": schema.StringAttribute{
+				Computed:    true,
+				Description: "Terraform resource ID (combination of organization_id, cluster_id, edge_location_id, and configuration_id)",
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.UseStateForUnknown(),
+				},
+			},
```
varnastadeus, line 60: why do we need this?

castai/resource_edge_configuration_default.go
```diff
@@ -0,0 +1,321 @@
...
+	}
+
+	// Update computed fields from the API response
+	state.Name = types.StringValue(apiResp.JSON200.Name)
+	state.CloudProvider = r.getCloudProviderType(apiResp.JSON200)
+
+	resp.Diagnostics.Append(resp.State.Set(ctx, state)...)
+}
+
+func (r *edgeConfigurationDefaultResource) Update(ctx context.Context, req resource.UpdateRequest, resp *resource.UpdateResponse) {
```
varnastadeus, line 240: based on schema there is no way to update the resource, we dont need this logic

castai/data_source_edge_configurations.go
andrejatcastai: Edge configuration is created by the backend when creating a location even if everything is managed by TF. So data source allows picking it up by terraform like following (at least that was the intention):
```
import {
  to = castai_edge_configuration.default
  id = "${var.organization_id}/${var.cluster_id}/${castai_edge_location.this.id}/${data.castai_edge_configurations.default.configurations[0].id}"
}

resource "castai_edge_configuration" "default" {
  organization_id  = var.organization_id
  cluster_id       = var.cluster_id
  edge_location_id = castai_edge_location.this.id
  name             = "default"
  
  aws = {
    image_id = "ami-0abcdef1234567890"
    tags = {
      Environment = "production"
    }
  }
}
```

I can change from from ListEdgeConfig to GetEdgeConfig (singular from plural). Would you agree?

varnastadeus: can we allow importing by either id of edge configuration or name similar as in other places? that way we dont need data source at all we can use 
```
id = "${var.organization_id}/${var.cluster_id}/${castai_edge_location.this.id}/default"
```

also, if we are keeping data source: it is strange that we are returning list of edge configurations and picking [0], we cannot necessarily know that 0 is the default one we want

castai/resource_edge_configuration.go
andrejatcastai: Removed

castai/data_source_edge_configurations.go
andrejatcastai: OK, will investigate...

castai/resource_edge_configuration.go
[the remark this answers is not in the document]
andrejatcastai: Fixed

andrejatcastai: Removed

castai/resource_edge_configuration.go
andrejatcastai: Fixed

castai/resource_edge_configuration.go
andrejatcastai: Fixed

castai/resource_edge_configuration.go
andrejatcastai: Fixed

castai/resource_edge_configuration.go
andrejatcastai: Renamed

castai/resource_edge_configuration_default.go
andrejatcastai: Removed

castai/resource_edge_configuration.go
```diff
@@ -0,0 +1,735 @@
...
+		config.ImageId = lo.ToPtr(plan.ImageID.ValueString())
+	}
+
+	if !plan.BootDiskSizeGiB.IsNull() {
+		config.BootDiskSizeGib = lo.ToPtr(int32(plan.BootDiskSizeGiB.ValueInt64()))
+	}
+
+	if !plan.Labels.IsNull() {
+		labels := make(map[string]string)
+		plan.Labels.ElementsAs(context.Background(), &labels, false)
```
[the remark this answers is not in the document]
andrejatcastai: Fixed

castai/resource_edge_configuration.go
```diff
@@ -0,0 +1,735 @@
...
+	state := &edgeConfigurationModel{
+		ID:             types.StringValue(lo.FromPtr(config.Id)),
+		OrganizationID: organizationID,
+		ClusterID:      clusterID,
+		Name:           types.StringValue(config.Name),
+		EdgeLocationID: types.StringValue(lo.FromPtr(config.EdgeLocationId)),
+		Default:        types.BoolValue(lo.FromPtr(config.Default)),
+		UserDataBase64: types.StringValue(lo.FromPtr(config.UserDataBase64)),
+	}
+
```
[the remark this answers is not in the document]
andrejatcastai: Fixed

castai/resource_edge_configuration.go
andrejatcastai: Fixed

castai/resource_edge_configuration.go
andrejatcastai: This mimics our API proto where we have:
```
  // Boot disk size in GiB.
  optional int32 boot_disk_size_gib = 4;
```
In case of your scenario: what will happen is that after changing from 500 to nil, new edges will use a default values for storage:
```
func getBootDiskSizeGiB(requestedBootDiskGiB uint32, edgeConfig *entities.EdgeConfiguration) uint32 {
	if requestedBootDiskGiB > 0 {
		return requestedBootDiskGiB
	}

	if edgeConfig != nil {
		var bootDiskGiB uint32
		switch {
		case edgeConfig.Config.GCPEdgeConfig != nil:
			bootDiskGiB = lo.FromPtr(edgeConfig.Config.GCPEdgeConfig.BootDiskSizeGiB)
		case edgeConfig.Config.AWSEdgeConfig != nil:
			bootDiskGiB = lo.FromPtr(edgeConfig.Config.AWSEdgeConfig.BootDiskSizeGiB)
		case edgeConfig.Config.OCIEdgeConfig != nil:
			bootDiskGiB = lo.FromPtr(edgeConfig.Config.OCIEdgeConfig.BootDiskSizeGiB)
		}
		if bootDiskGiB > 0 {
			return bootDiskGiB
		}
	}

	return entities.DefaultEdgeConfigurationBootDiskSize
}
```

castai/resource_edge_configuration_default.go
andrejatcastai: You are right, using composit ID is excessive. Changed to configuration ID.

castai/data_source_edge_configurations.go
andrejatcastai: Implemented support for importing resources as you suggested. Now we support both: supporting by name and by ID.

castai/resource_edge_configuration.go
andrejatcastai: Tested this scenario: when field is defined in TF then it is stored in DB. When field is removed in TF then it is removed in DB. So basically TF definition reflects what we have in DB.

castai/data_source_edge_configuration.go
```diff
@@ -0,0 +1,332 @@
...
+			"user_data_base64": schema.StringAttribute{
+				Computed:    true,
+				Description: "Base64 encoded user data for edge bootstrap",
+			},
+			"aws": schema.SingleNestedAttribute{
+				Computed:    true,
+				Description: "AWS specific configuration",
+				Attributes: map[string]schema.Attribute{
+					"image_id": schema.StringAttribute{
+						Computed:    true,
```
varnastadeus, line 104: Why are all fields computed?

andrejatcastai: This is a data source, so most fields (including nested) are Computed. This is inline with other data sources, for example we have this for Omni cluster:
```
			"castai_oidc_config": schema.SingleNestedAttribute{
				Computed:    true,
				Description: "CAST AI OIDC configuration for service account impersonation",
				Attributes: map[string]schema.Attribute{
					"gcp_service_account_email": schema.StringAttribute{
						Computed:    true,
						Description: "CAST AI GCP service account email for impersonation",
					},
					"gcp_service_account_unique_id": schema.StringAttribute{
						Computed:    true,
						Description: "CAST AI GCP service account unique ID for impersonation",
					},
				},
			},
```

varnastadeus: my bad, missed its a datasource. Do we still need it btw?

castai/data_source_edge_configuration.go
```diff
@@ -0,0 +1,350 @@
...
+		data.Custom = customConfigToTF(config.Custom)
+	} else {
+		data.Custom = types.MapNull(types.StringType)
+	}
+
+	resp.Diagnostics.Append(resp.State.Set(ctx, &data)...)
+}
+
+// int32ToInt64 converts *int32 to types.Int64
+func int32ToInt64(v *int32) types.Int64 {
```
varnastadeus: cant we use built in types.Int64PointerValue?

castai/data_source_edge_configuration.go
```diff
@@ -0,0 +1,350 @@
...
+	if v == nil {
+		return types.Int64Null()
+	}
+	return types.Int64Value(int64(*v))
+}
+
+// stringMapToTF converts map[string]string to types.Map
+func stringMapToTF(m *map[string]string) types.Map {
```
varnastadeus: I dont think you need this helper, all you need to do is nil check and use types.MapValueFrom

castai/data_source_edge_configuration.go
```diff
@@ -0,0 +1,350 @@
...
+
+	result, diags := types.MapValue(types.StringType, elements)
+	if diags.HasError() {
+		return types.MapNull(types.StringType)
+	}
+	return result
+}
+
+// customConfigToTF converts map[string]interface{} to types.Map
+func customConfigToTF(m *omni.CustomCloudConfiguration) types.Map {
```
varnastadeus: why do we need so much logic to map empty object? Also, why not do the same as in aws/gcp/oci and create empty model object and use it instead

castai/data_source_edge_configuration.go
andrejatcastai: Not sure I follow: this helper is used for boot disk size, which is defined as:
```
BootDiskSizeGiB types.Int64  `tfsdk:"boot_disk_size_gib"`
```

It is not a pointer, so not sure what you mean by using types.Int64PointerValue?

castai/resource_edge_configuration.go
```diff
@@ -0,0 +1,856 @@
...
+	resp.Diagnostics.Append(d...)
+
+	ociConfig, d := r.toOCIConfiguration(ctx, plan.OCI)
+	resp.Diagnostics.Append(d...)
+
+	customConfig, d := r.toCustomConfiguration(ctx, plan.Custom)
+	resp.Diagnostics.Append(d...)
+
+	criConfig, d := r.toCRIConfiguration(ctx, plan.CRI)
+	resp.Diagnostics.Append(d...)
```
varnastadeus, line 411: we need to able to reset this configuration, now if user had it set and then removes it, we wont send anything in the update request

castai/data_source_edge_configuration.go
andrejatcastai: I refactored this part to get rid of int32ToInt64 helper

castai/data_source_edge_configuration.go
andrejatcastai: Removed this helper

castai/resource_edge_configuration.go
andrejatcastai: testing this scenario

castai/data_source_edge_configuration.go
andrejatcastai: Removed this in the latest refactoring

castai/resource_edge_configuration.go
andrejatcastai: I was able to reproduce this issue. Made a change that handles it: https://github.com/castai/terraform-provider-castai/pull/700/changes/e8efe6387ec533868362926c43c96551d003972e

Basically if user removes CRI block and it is null in the state then we still send an empty block to the backend. This approach will be aligned with the module, it will always send CRI block as well (that's why I never hit this problme, it pops up only when using resource directly).

castai/data_source_edge_configuration.go
```diff
@@ -0,0 +1,331 @@
...
+		if config.Gcp.BootDiskSizeGib != nil {
+			data.Gcp.BootDiskSizeGiB = types.Int64Value(int64(*config.Gcp.BootDiskSizeGib))
+		}
+		if config.Gcp.Labels != nil && len(*config.Gcp.Labels) > 0 {
+			labels, diags := types.MapValueFrom(ctx, types.StringType, *config.Gcp.Labels)
+			if !diags.HasError() {
+				data.Gcp.Labels = labels
+			} else {
+				resp.Diagnostics.Append(diags...)
+			}
```
varnastadeus: if you introduce `var diags diag.Diagnostics` above there is no need to check error all the time and you can assign result straight away
```suggestion
			data.Gcp.Labels, diags = types.MapValueFrom(ctx, types.StringType, *config.Gcp.Labels)
```

Also, before setting state you should return if there are errors which now is not handled
```go
	if diags.HasError() {
		return
	}
```

varnastadeus (approved):

castai/data_source_edge_configuration.go
andrejatcastai: Fixed as suggested

PR #705, opened 2026-05-22
suppress diff in duration field e.g. 60s same as 1m

furkhat:
---
### What changed
Adds a `diffSuppressDuration` helper and applies it to string-typed duration fields in the autoscaler and AI optimizer hosted model resources so that equivalent durations in different units no longer trigger plan diffs. Also updates the generated SDK to include GPU workload optimization settings.

### Why
Terraform was detecting changes when the API returned durations using different units than those specified in configuration (e.g., `1m` versus `60s`), causing noisy and unnecessary resource updates.

### Key changes
- `castai/utils.go`: Added `diffSuppressDuration` helper that parses duration strings via `time.ParseDuration` and suppresses diffs when old and new values represent the same duration; falls back to string equality for unparseable values.
- `castai/resource_autoscaler.go`: Added `DiffSuppressFunc: diffSuppressDuration` to `FieldEvictorCycleInterval` and `FieldEvictorPodEvictionFailureBackOffInterval`.
- `castai/resource_ai_optimizer_hosted_model.go`: Added `DiffSuppressFunc: diffSuppressDuration` to `fieldAIHostedModelConditionDuration`.
- `castai/sdk/api.gen.go`: Added `WorkloadoptimizationV1GPUSettings` type and the `Gpu` field to `WorkloadoptimizationV1RecommendationPolicies`.
- `castai/utils_test.go`: Added `TestDiffSuppressDuration` covering equivalent durations across units, non-equivalent durations, invalid inputs, and empty strings.

### DIFF
--- a/castai/resource_autoscaler.go
+++ b/castai/resource_autoscaler.go
@@ -468,10 +468,11 @@ func resourceAutoscaler() *schema.Resource {
 													Description: "enable/disable scoped mode. By default, Evictor targets all nodes in the cluster. This mode will constrain it to just the nodes which were created by CAST AI.",
 												},
 												FieldEvictorCycleInterval: {
-													Type:        schema.TypeString,
-													Optional:    true,
-													Default:     "1m",
-													Description: "configure the interval duration between Evictor operations. This property can be used to lower or raise the frequency of the Evictor's find-and-drain operations.",
+													Type:             schema.TypeString,
+													Optional:         true,
+													Default:          "1m",
+													DiffSuppressFunc: diffSuppressDuration,
+													Description:      "configure the interval duration between Evictor operations. This property can be used to lower or raise the frequency of the Evictor's find-and-drain operations.",
 												},
 												FieldEvictorNodeGracePeriodMinutes: {
 													Type:        schema.TypeInt,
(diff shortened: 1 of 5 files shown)


castai/resource_autoscaler.go
```diff
@@ -1042,6 +1044,15 @@ func createValidationError(field, value string) error {
 		"Policy:\n%v", field, value)
 }
 
+func diffSuppressDuration(k, oldValue, newValue string, d *schema.ResourceData) bool {
```
Trojan295: I would either add a comment or rename the functions, so it's clear that this function checks for equality of provided durations, as in Go `1m` and `60s` are the same.

dvoros-castai: I think the name is okay as this is a [SchemaDiffSuppressFunc](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema#SchemaDiffSuppressFunc) function, but some comments would help. Also maybe put it under `utils.go` as it could be used for other resources too?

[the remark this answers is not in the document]
dvoros-castai: I agree, some unit tests would be easy enough to add. 🙏

PR #706, opened 2026-05-22
PRICE-1653: Expose pod eviction flag for Pod Mutations

michal-kopczynski:
---
### What changed
Adds a `pod_eviction` block to the `castai_pod_mutation` resource so users can opt in to evicting non-conforming pods when enforcing mutations.

### Why
This exposes the eviction enforcement feature in the provider, allowing mutations that require specific node constraints (for example, spot-only) to evict non-compliant pods and force them to be rescheduled onto suitable nodes.

### Key changes
- `castai/resource_pod_mutation.go`:
  - Added `pod_eviction` optional block (max 1 item) containing a required `enabled` boolean.
  - Updated `stateToPodMutation` to map the `pod_eviction` block to the API `PodEviction` struct.
  - Updated `podMutationToState` to normalize the API default of `Enabled=false` to an empty list, preventing perpetual state diffs.
- `castai/resource_pod_mutation_test.go`: Added unit tests for `pod_eviction` serialization, deserialization, and the `Enabled=false` normalization path.
- `docs/resources/pod_mutation.md` and `examples/resources/castai_pod_mutation/resource.tf`: Added schema documentation and an example resource demonstrating eviction enforcement.

### Impact
- **State normalization**: When the API returns `Enabled=false` (default), Terraform stores an empty list in state instead of an explicit `enabled = false` block. Users who want eviction active must explicitly set `pod_eviction { enabled = true }`.
- This is an additive, backward-compatible change.


castai/resource_pod_mutation.go
```diff
@@ -1277,6 +1302,20 @@ func podMutationToState(mutation *patching_engine.PodMutation, d *schema.Resourc
 		}
 	}
 
+	// Pod eviction — normalize API default (Enabled=false) to empty list to avoid perpetual diff
+	if mutation.PodEviction != nil && mutation.PodEviction.Enabled != nil && *mutation.PodEviction.Enabled {
+		evictionMap := map[string]interface{}{
+			"enabled": lo.FromPtr(mutation.PodEviction.Enabled),
```
maxtwardowski, line 1308: Only now I noticed that the pod mutation resource doesn't use `mapstructure` struct tags like commitments for example 😢 This would make the code way cleaner. Anyway, something to refactor in the future

maxtwardowski (approved):

PR #709, opened 2026-05-28
fix: omni edge-location validation

cheina97:
This pull request refactors how the `region` field is validated for `aws`, `gcp`, and `oci` edge location configurations in the `edge_location` Terraform resource. Instead of using a custom resource validator, it now leverages built-in object validators directly on the schema attributes, simplifying the code and making validation more declarative.

**Validation improvements:**

* The custom `edgeLocationRegionValidator` was removed, reducing custom code and potential maintenance overhead. [[1]](diffhunk://#diff-18fcc45190a4cb0295d893a130c324dd502c89e91c1959713a8ffd38dbf74f20L203-L227) [[2]](diffhunk://#diff-18fcc45190a4cb0295d893a130c324dd502c89e91c1959713a8ffd38dbf74f20L236)
* The `objectvalidator.AlsoRequires(path.MatchRoot("region"))` validator was added to the `aws`, `gcp`, and `oci` nested attributes in the schema, ensuring that `region` is required when any of these providers are configured. [[1]](diffhunk://#diff-18fcc45190a4cb0295d893a130c324dd502c89e91c1959713a8ffd38dbf74f20R325-R327) [[2]](diffhunk://#diff-18fcc45190a4cb0295d893a130c324dd502c89e91c1959713a8ffd38dbf74f20R384-R386) [[3]](diffhunk://#diff-18fcc45190a4cb0295d893a130c324dd502c89e91c1959713a8ffd38dbf74f20R431-R433)

**Dependency updates:**

* The `objectvalidator` package was imported to support the new validation approach.

---
### What changed
Fixes plan-time validation failures in the `castai_edge_location` resource by replacing a custom config validator with native schema-level `objectvalidator.AlsoRequires` rules. Also adds `useCastctl` query parameter support to the external cluster connect endpoint and introduces unit tests for validation behavior.

### Why
The previous `edgeLocationRegionValidator` decoded the entire resource config into a Go struct, which fails when Terraform marks any attribute as unknown (e.g., `zones` or `organization_id` referencing another resource's output). Schema-level validators only inspect the specific attributes they constrain and correctly tolerate unknown values.

### Key changes
- `castai/resource_edge_location.go`: Removed `edgeLocationRegionValidator` and its registration in `ConfigValidators`. Added `objectvalidator.AlsoRequires(path.MatchRoot("region"))` validators to the `aws`, `gcp`, and `oci` schema attributes.
- `castai/resource_edge_location_validate_test.go`: Added tests verifying that `aws`, `gcp`, and `oci` require `region`, that `custom` does not, and a regression test ensuring unknown config values no longer trigger spurious validation errors.
- `castai/sdk/api.gen.go`: Added `UseCastctl` boolean field to `ExternalClusterAPIGetConnectAndEnableCASTAICmdParams`.
- `castai/sdk/client.gen.go`: Added query-string serialization for the `useCastctl` parameter.

### Impact
- Resolves plan-time crashes when `edge_location` contains unknown values such as interpolated `zones`, `organization_id`, or `cluster_id`.

### DIFF
--- a/castai/resource_edge_location_validate_test.go
+++ b/castai/resource_edge_location_validate_test.go
@@ -1,0 +1,244 @@
+package castai
+
+import (
+	"context"
+	"testing"
+
+	"github.com/hashicorp/terraform-plugin-framework/path"
+	fwresource "github.com/hashicorp/terraform-plugin-framework/resource"
+	rschema "github.com/hashicorp/terraform-plugin-framework/resource/schema"
+	"github.com/hashicorp/terraform-plugin-framework/schema/validator"
+	"github.com/hashicorp/terraform-plugin-framework/tfsdk"
+	"github.com/hashicorp/terraform-plugin-framework/types"
+	"github.com/hashicorp/terraform-plugin-go/tftypes"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// edgeLocationSchema returns the schema for the edge location resource.
+func edgeLocationSchema(t *testing.T) fwresource.SchemaResponse {
+	t.Helper()
+	ctx := context.Background()
+	schemaResp := fwresource.SchemaResponse{}
+	newEdgeLocationResource().Schema(ctx, fwresource.SchemaRequest{}, &schemaResp)
+	return schemaResp
+}
+
+// edgeLocationSchemaObj returns the tftypes.Object representation of the edge
+// location schema, used to build typed tftypes.Value instances in tests.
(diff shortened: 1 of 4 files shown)


castai/resource_edge_location_validate_test.go
```diff
@@ -0,0 +1,244 @@
+package castai
```
varnastadeus, line 1: I dont think we need this file as we are not testing any new logic apart of built in framework helpers that are tested on framework side

varnastadeus:
closing in favor of #712

PR #713, opened 2026-06-01
fix: cri conversion in case of empty values

andrejatcastai:
---
### What changed  
Fix spurious diff detection in the `castai_edge_configuration` resource when the API returns a CRI configuration object with an empty or absent `socket` value. State is now normalized after `Create` and `Update` to align empty CRI blocks with the Terraform plan.

### Why  
When the API returns a non-nil CRI object containing an empty or missing socket, Terraform incorrectly reports a diff against configurations that omit the CRI block entirely. This change prevents perpetual configuration drift by treating empty/absent sockets as null and dropping the CRI state when the plan leaves it unspecified.

### Key changes  
- `castai/resource_edge_configuration.go`: Added `normalizeCRIState` helper invoked after `Create` and `Update` to convert empty-string CRI sockets to `types.StringNull()` and clear `state.CRI` to `nil` when the plan omits the CRI block.  
- `castai/resource_edge_configuration.go`: Updated `toCRIConfigurationModel` to return a non-nil `criConfigurationModel` with a null `Socket` instead of returning `nil` when the API omits the socket value, preserving structural parity before normalization.

### DIFF
--- a/castai/resource_edge_configuration.go
+++ b/castai/resource_edge_configuration.go
@@ -317,6 +317,17 @@ func (r *edgeConfigurationResource) Create(ctx context.Context, req resource.Cre
 	state := r.edgeConfigurationToTFModel(ctx, apiResp.JSON200, plan.OrganizationID, plan.ClusterID)
 	state.EdgeLocationID = plan.EdgeLocationID
 
+	// Normalize CRI state to prevent spurious diffs when the API returns
+	// a non-nil CRI object with an empty string socket.
+	if state.CRI != nil && state.CRI.Socket.ValueString() == "" {
+		state.CRI.Socket = types.StringNull()
+	}
+
+	// if the entire cri block was removed; match by returning nil.
+	if plan.CRI == nil {
+		state.CRI = nil
+	}
+
 	resp.Diagnostics.Append(resp.State.Set(ctx, state)...)
 }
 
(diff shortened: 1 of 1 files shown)


castai/resource_edge_configuration.go
```diff
@@ -317,6 +317,17 @@ func (r *edgeConfigurationResource) Create(ctx context.Context, req resource.Cre
 	state := r.edgeConfigurationToTFModel(ctx, apiResp.JSON200, plan.OrganizationID, plan.ClusterID)
 	state.EdgeLocationID = plan.EdgeLocationID
 
+	// Normalize CRI state to prevent spurious diffs when the API returns
+	// a non-nil CRI object with an empty string socket.
+	if state.CRI != nil && state.CRI.Socket.ValueString() == "" {
```
anjmao: This is duplicate in two places, is it possible to have this inside `edgeConfigurationToTFModel` method?

andrejatcastai: No, because it does not have access to a plan. Moreover it is used only for Create and Update. It is irrelevant for Read. But I can create a helper and reuse.

andrejatcastai: Would you agree?
https://github.com/castai/terraform-provider-castai/pull/713/changes/c61ff8c77155d181659afa3105d9215cfd0f4cef

anjmao (approved):

PR #725, opened 2026-06-16
fix: live set gp2 as default storage class

Mitiaj:
---
### What changed
The EKS live migration example now waits for the `aws-ebs-csi-driver` addon to reconcile, then patches the EKS-created `gp2` StorageClass with the default annotation. This allows PVCs without an explicit `storageClassName` to provision volumes successfully.

### Why
EKS creates the `gp2` StorageClass asynchronously after cluster creation and does not mark it as the default storage class. Without this annotation, workloads that rely on a default StorageClass fail to bind persistent volume claims.

### Key changes
- `examples/eks/eks_cluster_live_migration/eks.tf`: Added `time_sleep.wait_for_gp2` to pause 30 seconds after `module.eks` completes, allowing the addon to create the `gp2` StorageClass; added `kubernetes_annotations.gp2_default_storage_class` to apply the `storageclass.kubernetes.io/is-default-class` annotation to `gp2`.
- `examples/eks/eks_cluster_live_migration/versions.tf`: Added the `hashicorp/time` provider with version constraint `~> 0.9`.

### DIFF
--- a/examples/eks/eks_cluster_live_migration/eks.tf
+++ b/examples/eks/eks_cluster_live_migration/eks.tf
@@ -95,6 +95,22 @@ module "ebs_csi_irsa_role" {
   }
 }
 
+# Mark the EKS-created gp2 StorageClass as the cluster default.
+# EKS ships gp2 without the default annotation, so PVCs without an explicit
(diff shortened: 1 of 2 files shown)


Mitiaj:
@kimchi review

PR #730, opened 2026-06-24
CSU-5514: expose synchronize_user_groups field for sso_connection resource

radekska:
---
### What changed
Adds `synchronize_user_groups` support to the `castai_sso_connection` resource, allowing Terraform to enable identity-provider user group synchronization with CAST AI. When activated, a one-time sync auth token is captured in the computed `sync_auth_token` attribute. Also updates the `Makefile` default target to print a documented `help` message instead of building immediately.

### Why
Enables SSO connections managed via Terraform to opt into user group synchronization and surface the required one-time auth token for downstream configuration.

### Key changes
- `castai/resource_sso_connection.go`
  - Added `synchronize_user_groups` (bool, optional, default `false`) and `sync_auth_token` (string, computed, sensitive) schema fields.
  - `resourceCastaiSSOConnectionCreate`: calls `setSSOConnectionSync` to invoke `SSOAPISetSyncForSSOConnectionWithResponse` when `synchronize_user_groups` is `true` on creation.
  - `resourceCastaiSSOConnectionRead`: maps the API `isSynced` field to `synchronize_user_groups` and explicitly preserves `sync_auth_token` from existing state since the GET endpoint never returns it.
  - `resourceCastaiSSOConnectionUpdate`: splits base connection updates from sync-toggle updates; calls SetSync only on actual `synchronize_user_groups` transitions.
  - Added `setSSOConnectionSync` helper to store the returned token and emit a warning diagnostic so users know to save it immediately.
- `castai/resource_sso_connection_test.go`: added unit tests for create/read with sync enabled and disabled, token persistence, and `setSSOConnectionSync` behavior.
- `castai/sdk/patching_engine/api.gen.go`: added `ExcludeTolerations` field to `ObjectFilterV2`.
- `docs/resources/sso_connection.md`: documented `synchronize_user_groups` and `sync_auth_token`.
- `Makefile`: changed default target to `help` and documented all available targets.

### Impact
- `sync_auth_token` is a one-time secret returned only when `synchronize_user_groups` transitions from `false` to `true`. It must be captured from Terraform state or outputs immediately; subsequent refreshes do not return it from the API.
- Running `make` without arguments now prints help text instead of building the provider.

### DIFF
--- a/castai/resource_sso_connection.go
+++ b/castai/resource_sso_connection.go
@@ -237,6 +237,13 @@ func resourceCastaiSSOConnectionRead(ctx context.Context, data *schema.ResourceD
 		return diag.Errorf("setting synchronize_user_groups: %v", err)
 	}
 
+	// sync_auth_token is a one-time value returned only by the SetSync endpoint — the GET
+	// API never returns it. Explicitly write back the current state value so Terraform does
+	// not drop it from state during a refresh.
+	if err := data.Set(FieldSSOConnectionSyncAuthToken, data.Get(FieldSSOConnectionSyncAuthToken)); err != nil {
+		return diag.Errorf("setting sync_auth_token: %v", err)
+	}
+
 	return nil
 }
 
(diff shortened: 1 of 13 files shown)


castai/resource_sso_connection.go
```diff
@@ -340,3 +394,37 @@ func toOktaConnector(obj map[string]any) *sdk.CastaiSsoV1beta1Okta {
...
+			{
+				Severity: diag.Warning,
+				Summary:  "SSO sync auth token generated",
+				Detail:   "A sync auth token was generated and stored in sync_auth_token. Retrieve it from the Terraform state or outputs now — it is only returned once when synchronization is enabled.",
+			},
+		}
+	}
+
+	// On disable, clear the stored token.
+	if err := data.Set(FieldSSOConnectionSyncAuthToken, ""); err != nil {
```
oanatmaria: we should keep it. I think for re-enablement we reuse the token

oanatmaria (approved):

radekska:
@getkimchi review

PR #734, opened 2026-07-01
CID-199 enterprise service accounts

plks:
Adds `castai_enterprise_service_account resource` for managing service accounts within a CAST AI enterprise organization, either in the enterprise org itself or in any of its child orgs.

---
### What changed

Introduces a new Terraform resource, `castai_enterprise_service_account`, to manage CAST AI enterprise service accounts through the provider. It supports full CRUD lifecycle operations, with paginated reads and batch API calls for create, update, and delete.

### Why

Enterprise users need to automate service account provisioning and lifecycle management across their organizations via Terraform, avoiding manual configuration in the CAST AI console.

### Key changes

- `castai/provider.go`: Registers the new `castai_enterprise_service_account` resource in the provider.
- `castai/resource_enterprise_service_account.go`: Implements the resource with schema fields `enterprise_id`, `organization_id`, `name`, `description`, and computed `email`. Both `enterprise_id` and `organization_id` are marked `ForceNew`. Uses `EnterpriseAPIBatchCreateEnterpriseServiceAccountsWithResponse`, `EnterpriseAPIBatchUpdateEnterpriseServiceAccountsWithResponse`, and `EnterpriseAPIBatchDeleteEnterpriseServiceAccountsWithResponse` for mutations, and paginated `EnterpriseAPIListEnterpriseServiceAccountsWithResponse` for reads.
- `castai/resource_enterprise_service_account_test.go`: Adds unit tests for create, read, update, and delete contexts, covering success paths, API errors, empty responses, and pagination.
- `docs/resources/enterprise_service_account.md`: Adds generated resource documentation.
- `examples/enterprise_service_accounts/main.tf` and `examples/resources/castai_enterprise_service_account/resource.tf`: Adds example Terraform configurations.


the-marian (approved):

examples/resources/castai_enterprise_service_account/resource.tf
oskarwojciski, line 1: I assume that those two examples are:
1. Create service account with child org scope access
2. Create service account with enterprise org scope access?

Can you add description/comments into the examples?
Also does it makes sens to have them on two paths? Isn't the `resources` enough?

examples/resources/castai_enterprise_service_account/resource.tf
```diff
@@ -0,0 +1,6 @@
+resource "castai_enterprise_service_account" "example" {
+  enterprise_id   = "3e81b0c1-ea70-4513-8c11-60260fa04fba"
+  organization_id = "3e81b0c1-ea70-4513-8c11-60260fa04fba"
```
oskarwojciski: I am wondering - maybe we can have a logic that if the organization_id is not provided we create the service account  with enterprise scope by default? (in the provider it is set that organization_id = enterprise_id)

oskarwojciski (approved):

PR #738, opened 2026-07-02
Add readonly EKS example with kvisor

jakubsolecki:
The example covers Kvisor with storage metrics enabled, but can be expanded to include other monitoring features.

---
### What changed
Adds a new Terraform example under `examples/eks/eks_cluster_readonly_kvisor/` that demonstrates how to onboard an existing EKS cluster to CAST AI in read-only mode and install the CAST AI Kvisor component to collect storage metrics. Also adds `.kimchi/` to `.gitignore`.

### Why
Provides a ready-to-use reference configuration for users who want to run CAST AI in read-only mode (Phase 1) while enabling Kvisor storage metrics collection via IRSA.

### Key changes
- **`examples/eks/eks_cluster_readonly_kvisor/castai.tf`** — Configures `castai_eks_cluster` for read-only onboarding, deploys the `castai-agent` Helm chart, creates an IRSA IAM role with `ec2:DescribeVolumes` permission, and deploys the `castai-kvisor` Helm chart with storage metrics enabled.
- **`examples/eks/eks_cluster_readonly_kvisor/providers.tf`** — Configures the `aws`, `castai`, `kubernetes`, and `helm` providers, including EKS `get-token` exec authentication.
- **`examples/eks/eks_cluster_readonly_kvisor/variables.tf`** — Defines input variables for `cluster_name`, `cluster_region`, `castai_api_token`, `castai_api_url`, and `kvisor_grpc_addr`.
- **`examples/eks/eks_cluster_readonly_kvisor/versions.tf`** — Pins required provider versions (`helm ~> 3.0`, `aws ~> 6.0`) and sets Terraform `>= 1.3.2`.
- **`examples/eks/eks_cluster_readonly_kvisor/terraform.tfvars.example`** — Provides placeholder values for required variables.
- **`examples/eks/eks_cluster_readonly_kvisor/README.MD`** — Documents prerequisites, configuration order, usage steps, and variable descriptions.
- **`.gitignore`** — Adds `.kimchi/` to the ignore list.

### Impact
No breaking changes. These are additive documentation/example files plus a minor `.gitignore` update.

### DIFF
--- a/examples/eks/eks_cluster_readonly_kvisor/castai.tf
+++ b/examples/eks/eks_cluster_readonly_kvisor/castai.tf
@@ -1,0 +1,155 @@
+# Connect existing EKS cluster to CAST AI in read-only mode and install Kvisor for storage metrics.
+
+data "aws_caller_identity" "current" {}
+
+data "aws_eks_cluster" "existing_cluster" {
+  name = var.cluster_name
+}
+
+data "aws_iam_openid_connect_provider" "eks" {
+  url = data.aws_eks_cluster.existing_cluster.identity[0].oidc[0].issuer
+}
+
+locals {
+  oidc_issuer = trimprefix(data.aws_eks_cluster.existing_cluster.identity[0].oidc[0].issuer, "https://")
+}
+
+# Configure EKS cluster connection to CAST AI in read-only mode.
+resource "castai_eks_cluster" "this" {
+  account_id = data.aws_caller_identity.current.account_id
+  region     = var.cluster_region
+  name       = var.cluster_name
+}
+
+resource "helm_release" "castai_agent" {
+  name             = "castai-agent"
+  repository       = "https://castai.github.io/helm-charts"
+  chart            = "castai-agent"
+  namespace        = "castai-agent"
+  create_namespace = true
+  cleanup_on_fail  = true
+
+  set = concat(
+    [
+      {
+        name  = "provider"
+        value = "eks"
+      },
+      {
+        # Required until https://github.com/castai/helm-charts/issues/135 is fixed.
+        name  = "createNamespace"
+        value = "false"
+      },
+    ],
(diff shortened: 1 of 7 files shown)


examples/eks/eks_cluster_readonly_kvisor/castai.tf
```diff
@@ -0,0 +1,155 @@
+# Connect existing EKS cluster to CAST AI in read-only mode and install Kvisor for storage metrics.
```
RomanMelnyk113, line 1: I'm thinking as this is named as general kvisor example, but here we specify that we enabling storage metrics, maybe we then should rename example title to explicitly set storage in name? 

also based on our naming pattern if we onboard existing EKS cluster it should contain `existing` word 

i.e. `examples/eks/eks_cluster_existing_kvisor_storage_metrics/castai.tf`

jakubsolecki: I honestly really don't like the pattern that we have 1000 examples, one for each small scenario. In my eyes, so many examples devalue each other. I was even thinking of including Kvisor in the existing read-only example. My idea for this one is to make it extensible, so later more Kvisor monitoring features could be included here (GPU, netflow, reliability, whatever is decided as needed/encouraged for even basic monitoring)

jakubsolecki: But if you think we should just follow how things were done already, sure, I don't mind changing example name

RomanMelnyk113: depends on use case, I find it handy to have specific examples per feature even if it is duplicates a lot from another example, but overall I'm fine having it like this, as making it extensible in the future makes sense for sure

RomanMelnyk113 (approved):

PR #739, opened 2026-07-02
CO-4590: Deprecate Node AWS Rebalancing Recommendation

nkotikov-castai:
This MR deprecates AWS Rebalancing Recommendation option `aws-rebalance-recommendations` for Resource Node Template, making Cast AI ML the only option: `interruption-predictions`.

---
### What changed
Deprecates the legacy AWS rebalance-recommendation spot interruption prediction type and makes Cast AI ML predictions the default across the provider. Existing configurations using the old value continue to work but are normalized internally, and Terraform diffs are suppressed when the old and new values map to the same underlying prediction type.

### Why
Cast AI ML predictions are now used for all spot interruption prediction, so the legacy `AWSRebalanceRecommendations` / `aws-rebalance-recommendations` option is obsolete. Smoothing the transition avoids forcing immediate state migrations while guiding users to the supported setting.

### Key changes

- `castai/resource_autoscaler.go`
  - Changed the default for `spot_interruption_predictions_type` from `AWSRebalanceRecommendations` to `CASTAIInterruptionPredictions`.
  - Marked the old value as deprecated.
  - Added `DiffSuppressFunc` to treat `AWSRebalanceRecommendations` as equivalent to `CASTAIInterruptionPredictions`.

- `castai/resource_node_template.go`
  - Added a default value of `interruption-predictions` for `FieldNodeTemplateSpotInterruptionPredictionsType`.
  - Deprecated `aws-rebalance-recommendations` and updated the description.
  - Added `DiffSuppressFunc` to suppress diffs between the deprecated and normalized values.
  - Updated `flattenConstraints` and `toTemplateConstraints` to normalize the deprecated value via `normalizeSpotInterruptionPredictionsType`.
  - Added `normalizeSpotInterruptionPredictionsType` helper.

- `castai/resource_node_template_test.go`
  - Added unit tests for normalization, read/write path normalization, and diff suppression behavior.

- `docs/resources/autoscaler.md` and `docs/resources/node_template.md`
  - Updated documentation to reflect deprecation and the new default.

### Impact

- Existing Terraform configs that explicitly set the deprecated value will continue to plan/apply cleanly because the diff suppression and normalization treat both values as equivalent.
- The effective default prediction type changes to Cast AI ML predictions (`CASTAIInterruptionPredictions` / `interruption-predictations`). For most users this is transparent, but plans may show a one-time suppressed update if the previous default was recorded.
- The deprecated values remain valid in `ValidateDiagFunc` for backward compatibility; they will be removed in a future major version.

### DIFF
--- a/castai/resource_autoscaler.go
+++ b/castai/resource_autoscaler.go
@@ -391,9 +391,19 @@ func resourceAutoscaler() *schema.Resource {
 												FieldSpotInterruptionPredictionsType: {
 													Type:             schema.TypeString,
 													Optional:         true,
-													Default:          "AWSRebalanceRecommendations",
-													Description:      "define the type of the spot interruption prediction to handle. Allowed values are AWSRebalanceRecommendations, CASTAIInterruptionPredictions.",
+													Default:          "CASTAIInterruptionPredictions",
+													Description:      "define the type of the spot interruption prediction to handle. The value \"AWSRebalanceRecommendations\" is deprecated; use \"CASTAIInterruptionPredictions\".",
+													Deprecated:       "The value \"AWSRebalanceRecommendations\" is deprecated. Cast AI ML predictions are now used for all spot interruption prediction.",
 													ValidateDiagFunc: validation.ToDiagFunc(validation.StringInSlice([]string{"AWSRebalanceRecommendations", "CASTAIInterruptionPredictions"}, false)),
+													DiffSuppressFunc: func(k, oldVal, newVal string, d *schema.ResourceData) bool {
+														normalize := func(v string) string {
+															if v == "AWSRebalanceRecommendations" {
+																return "CASTAIInterruptionPredictions"
+															}
+															return v
+														}
+														return normalize(oldVal) == normalize(newVal)
+													},
 												},
 											},
 										},
--- a/castai/resource_node_template.go
+++ b/castai/resource_node_template.go
@@ -291,8 +291,13 @@ func resourceNodeTemplate() *schema.Resource {
 						FieldNodeTemplateSpotInterruptionPredictionsType: {
 							Type:             schema.TypeString,
 							Optional:         true,
-							Description:      "Spot interruption predictions type. Can be either \"aws-rebalance-recommendations\" or \"interruption-predictions\".",
+							Default:          "interruption-predictions",
+							Description:      "Spot interruption predictions type. Only \"interruption-predictions\" is supported.",
+							Deprecated:       "The value \"aws-rebalance-recommendations\" is deprecated and will be removed in a future major version. Cast AI ML predictions (\"interruption-predictions\") are now used for all spot interruption prediction.",
 							ValidateDiagFunc: validation.ToDiagFunc(validation.StringInSlice([]string{"aws-rebalance-recommendations", "interruption-predictions"}, false)),
+							DiffSuppressFunc: func(k, oldVal, newVal string, d *schema.ResourceData) bool {
+								return normalizeSpotInterruptionPredictionsType(oldVal) == normalizeSpotInterruptionPredictionsType(newVal)
+							},
 						},
 						FieldNodeTemplateMinCpu: {
 							Type:        schema.TypeInt,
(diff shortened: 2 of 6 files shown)


castai/resource_autoscaler.go
```diff
@@ -391,9 +391,19 @@ func resourceAutoscaler() *schema.Resource {
 												FieldSpotInterruptionPredictionsType: {
 													Type:             schema.TypeString,
 													Optional:         true,
-													Default:          "AWSRebalanceRecommendations",
-													Description:      "define the type of the spot interruption prediction to handle. Allowed values are AWSRebalanceRecommendations, CASTAIInterruptionPredictions.",
+													Default:          "CASTAIInterruptionPredictions",
+													Description:      "define the type of the spot interruption prediction to handle. The value \"AWSRebalanceRecommendations\" is deprecated; use \"CASTAIInterruptionPredictions\".",
+													Deprecated:       "The value \"AWSRebalanceRecommendations\" is deprecated. Cast AI ML predictions are now used for all spot interruption prediction.",
```
thomasstrang-cast, line 396: This whole block is deprecated but I geuss we could update the field anyway - see line 382

castai/resource_node_template.go
```diff
@@ -290,19 +289,19 @@ func resourceNodeTemplate() *schema.Resource {
...
 							Type:             schema.TypeString,
 							Optional:         true,
-							Description:      "Spot interruption predictions type. Can be either \"aws-rebalance-recommendations\" or \"interruption-predictions\".",
+							Default:          "interruption-predictions",
+							Description:      "Spot interruption predictions type. Only \"interruption-predictions\" is supported.",
+							Deprecated:       "The value \"aws-rebalance-recommendations\" is deprecated and will be removed in a future major version. Cast AI ML predictions (\"interruption-predictions\") are now used for all spot interruption prediction.",
 							ValidateDiagFunc: validation.ToDiagFunc(validation.StringInSlice([]string{"aws-rebalance-recommendations", "interruption-predictions"}, false)),
+							DiffSuppressFunc: func(k, oldVal, newVal string, d *schema.ResourceData) bool {
+								return normalizeSpotInterruptionPredictionsType(oldVal) == normalizeSpotInterruptionPredictionsType(newVal)
+							},
```
thomasstrang-cast, line 300: We can just suppress this entire field
```suggestion
							DiffSuppressFunc: func(k, oldVal, newVal string, d *schema.ResourceData) bool {
								return True
							},
```

castai/resource_node_template.go
```diff
@@ -1076,8 +1075,10 @@ func flattenConstraints(c *sdk.NodetemplatesV1TemplateConstraints) ([]map[string
 		out[FieldNodeTemplateSpotInterruptionPredictionsEnabled] = c.SpotInterruptionPredictionsEnabled
 	}
 	if c.SpotInterruptionPredictionsType != nil {
-		out[FieldNodeTemplateSpotInterruptionPredictionsType] = c.SpotInterruptionPredictionsType
+		normalized := normalizeSpotInterruptionPredictionsType(*c.SpotInterruptionPredictionsType)
+		out[FieldNodeTemplateSpotInterruptionPredictionsType] = &normalized
 	}
+
```
thomasstrang-cast, line 1088: can just delete this

castai/resource_node_template.go
```diff
@@ -1753,7 +1748,7 @@ func toTemplateConstraints(obj map[string]any) *sdk.NodetemplatesV1TemplateConst
 		out.SpotInterruptionPredictionsEnabled = toPtr(v)
 	}
 	if v, ok := obj[FieldNodeTemplateSpotInterruptionPredictionsType].(string); ok {
-		out.SpotInterruptionPredictionsType = toPtr(v)
+		out.SpotInterruptionPredictionsType = toPtr(normalizeSpotInterruptionPredictionsType(v))
 	}
```
thomasstrang-cast, line 1767: can just delete this

thomasstrang-cast (changes requested):
LGTM, suggestions are not blocking if you disagree

castai/resource_node_template.go
nkotikov-castai: Won't it break silently and hide the diff if we add some third value?

thomasstrang-cast: if someone puts 'foobar' then it will show no diff yeah. But it won't break, it just won't send the deprecated field to the backend, if we delete the addition of out[FieldNodeTemplateSpotInterruptionPredictionsType] below

thomasstrang-cast (approved):

castai/resource_node_template.go
nkotikov-castai: Idk, I suggest we still show the diff, because why not?... I feel that it's better to show it?  Or is it a bad practice to do?

nkotikov-castai:
@getkimchi review

PR #744, opened 2026-07-03
feat: edge location addon support

varnastadeus:
---
### What changed
Adds optional `addons` support to the `castai_edge_location` resource. The new `addons` block installs optional edge cluster addons (e.g., `nvidia-gpu-operator`, `nvidia-network-operator`, `nvidia-dra`, `oci-csi`) with optional Helm `values` encoded as JSON. When `addons` is omitted, it defaults to `[{ name = "nvidia-gpu-operator" }]`. Setting `addons = []` explicitly disables all optional addons.

### Why
Edge clusters now expose optional addons that were previously unavailable through Terraform. This change lets users declare, customize, and update those addons declaratively alongside the rest of the edge location configuration.

### Key changes
- `castai/resource_edge_location.go`
  - Added `addons` schema attribute with a static default of `nvidia-gpu-operator`.
  - Added `addonModel`, `defaultAddonsList`, `addonsEqual`, `toAddons`, and `toAddonsModel` to map Terraform state to/from the SDK.
  - Included `addons` in create, read, update, and diff detection for `edge_cluster_spec`.
  - Fixed empty-string handling for `description` and `gcp.instance_service_account` to avoid perpetual diffs.
- `castai/validators/validators.go`
  - Added `ValidJSON` string validator for `addons.values`.
- `castai/sdk/omni/api.gen.go`
  - Added `EdgeClusterAddon` type and `Addons` field on `EdgeClusterSpec`.
- `castai/sdk/api.gen.go`
  - Synced generated SDK: added `CountSoleTenant` to node template list stats and `InfiniteMemoryScalingSettings`.
- `castai/resource_edge_location_test.go`
  - Added acceptance test steps covering default addons, custom addons, explicit empty list, and block removal.
- `docs/resources/edge_location.md` and `examples/resources/castai_edge_location/resource.tf`
  - Updated documentation and examples to show default, custom, and no-addon configurations.

### Impact
- **Behavioral default:** New or refreshed `castai_edge_location` resources now default to installing the `nvidia-gpu-operator` addon unless `addons = []` is set.
- **First-run diff:** Existing state may show an `addons` diff after upgrading the provider; apply to accept the default or set `addons` explicitly.
- **Addon updates:** Changing `addons` triggers an edge cluster spec update.
- No breaking changes to required fields.


castai/resource_edge_location.go
```diff
@@ -58,6 +62,30 @@ type edgeLocationModel struct {
 	CredentialsRevision types.Int64 `tfsdk:"credentials_revision"`
 }
 
+type addonModel struct {
+	Name   types.String `tfsdk:"name"`
+	Values types.String `tfsdk:"values"`
+}
+
+func (m addonModel) Equal(other addonModel) bool {
+	return m.Name.Equal(other.Name) && m.Values.Equal(other.Values)
```
andrejatcastai, line 71: Values string is serialized values. So potentially different values strings may be equal. For example if values are yaml and you change elements order then they are still equal but string representing will be different.

andrejatcastai (approved):

castai/resource_edge_location.go
varnastadeus: jsonencode already handles this, this is only true if you would manually construct the json string and then switch places which not sure why you would ever need to do this

fra98 (approved):

PR #748, opened 2026-07-07
chore: isDefault is only true on default-by-castai template

primoz-hrovat:


thomasstrang-cast (changes requested):
If we are ignoring what the user sets as FieldNodeTemplateIsDefault then we should deprecate FieldNodeTemplateIsDefault  and add a comment stating that the name is being used to define which is default. 

What happens if a user creates a new template with the same name default-by-castai ? I have not checked but would geuss there is a constraint in the backend and no constraint in the terraform provider - might be good to add such a restriction if possible.

primoz-hrovat:
> If we are ignoring what the user sets as FieldNodeTemplateIsDefault then we should deprecate FieldNodeTemplateIsDefault and add a comment stating that the name is being used to define which is default.

I don't want to deprecate it. The limitation is described in the [docs](https://docs.cast.ai/docs/node-templates#more-specific-requirementsconstraints), and there is a feature request to allow a different node template to be default. Meaning we would then need to revert the deprecation.

> What happens if a user creates a new template with the same name default-by-castai ? I have not checked but would geuss there is a constraint in the backend and no constraint in the terraform provider - might be good to add such a restriction if possible.

This is in the code for "create node template" - it also has a special case for `default-by-castai` named template (it already exists.

```
// When creating a default node template, use PUT instead of POST because a default node template is automatically created in the background.
// Since name of the default node template is fixed, we can use it to identify the default node template. All other
// requests should be treated as regular node template creation requests to avoid conflicts for validation of the request.
```

thomasstrang-cast (approved):

thomasstrang-cast:
> > If we are ignoring what the user sets as FieldNodeTemplateIsDefault then we should deprecate FieldNodeTemplateIsDefault and add a comment stating that the name is being used to define which is default.
> 
> I don't want to deprecate it. The limitation is described in the [docs](https://docs.cast.ai/docs/node-templates#more-specific-requirementsconstraints), and there is a feature request to allow a different node template to be default. Meaning we would then need to revert the deprecation.
> 
> > What happens if a user creates a new template with the same name default-by-castai ? I have not checked but would geuss there is a constraint in the backend and no constraint in the terraform provider - might be good to add such a restriction if possible.
> 
> This is in the code for "create node template" - it also has a special case for `default-by-castai` named template (it already exists.
> 
> ```
> // When creating a default node template, use PUT instead of POST because a default node template is automatically created in the background.
> // Since name of the default node template is fixed, we can use it to identify the default node template. All other
> // requests should be treated as regular node template creation requests to avoid conflicts for validation of the request.
> ```

If there is a plan to update is_default to mark a node template as default in the future then fair enough we shouldn't deprecate. But I would still update the docs to make it clear that at this point in time the is_default config doesn't do anything and shouldn't be used. Which I see you have done, but could be more explicit.

My question about if a user makes a second node template with the default name was just a question, not wanting to block your work. I would expect this to happen if they are copy pasting configs around etc. At this point in time it would cause 2 default templates to get uploaded

primoz-hrovat:
Indeed, there would be a race condition OR potential conflict (409) if the user were to "duplicate" the `default-by-castai` node template.

For now, I have updated the field description to (hopefully) make it clearer.

PR #752, opened 2026-07-10
CO-4398: deprecate min cpu cluster level setting keep max cpu

thomasstrang-cast:
deprecates the cluster wide min cpu config, as it is not properly supported anymore and will be removed soon.


castai/resource_autoscaler.go
```diff
@@ -294,10 +294,14 @@ func resourceAutoscaler() *schema.Resource {
...
-													Type:             schema.TypeInt,
-													Optional:         true,
-													Default:          1,
-													Description:      "defines the minimum allowed amount of CPUs in the whole cluster.",
+													Type:        schema.TypeInt,
+													Optional:    true,
+													Default:     1,
+													Deprecated:  "This field is no longer enforced and will be removed in a future major version of the provider. Remove it from your configuration.",
+													Description: "defines the minimum allowed number of CPUs in the whole cluster. This is deprecated and no longer enforced by autoscaler.",
+													DiffSuppressFunc: func(k, oldValue, newValue string, d *schema.ResourceData) bool {
```
nkotikov-castai: Won't it make people annoyed with perpetual diff like it did with my previous aws changes?

castai/resource_autoscaler.go
```diff
@@ -294,10 +294,14 @@ func resourceAutoscaler() *schema.Resource {
...
+														return true
+													},
```
primoz-hrovat: Reword this, please. minCpu is deprecated, but for the time being, we'll keep the enforcer in the node downscaler. It's unsupported elsewhere

castai/resource_autoscaler.go
```diff
@@ -648,9 +652,6 @@ func flattenClusterLimits(clusterLimits map[string]interface{}) map[string]inter
 
 	if cpu, ok := clusterLimits["cpu"].(map[string]interface{}); ok {
 		cpuMap := make(map[string]interface{})
-		if v, ok := cpu["minCores"]; ok {
```
primoz-hrovat, line 651: The field should still be populated, just deprecated

primoz-hrovat (changes requested):

castai/resource_autoscaler.go
thomasstrang-cast: I believe it won't but I shall run some tests locally and make sure it doesn't ...

thomasstrang-cast: I ended up not changing the behavioru of the min cpu field - it will still get applied, and not get suppressed. So this change was reverted.

I am not sure what change was made (if any) to the backend for the spot_interruption_predictions_type , but I am seeing a failing acceptance test due to a diff on spot_interruption_predictions_type. Possible the backend is no longer returning a spot_interruption_predictions_type value? So I have added a suppress function to prevent any diff whatsoever for spot_interruption_predictions_type.

primoz-hrovat: @nikita-castai please check on the above so this PR can be free from such unrelated changes.

nkotikov-castai: @primoz-hrovat @thomasstrang-cast fixed
Regarding cloud-agnostic pipeline: not sure about this... any ideas? I see other people's pipelines also fail with this =( 
```
=== NAME  TestAccCloudAgnostic_ResourceCacheRule
    resource_cache_rule_test.go:366: Step 1/1 error: Error running apply: exit status 1
        
        Error: cache configuration with ID 8e1b31cd-95a2-4924-92f9-19de84dc43bc not found in cache group 0ceb8bec-7fd2-43df-90bb-7c60120b6e11
        
          with castai_cache_configuration.test,
          on terraform_plugin_test.tf line 17, in resource "castai_cache_configuration" "test":
          17: resource "castai_cache_configuration" "test" ***
        
--- FAIL: TestAccCloudAgnostic_ResourceCacheRule (7.57s)
```

primoz-hrovat (approved):

PR #754, opened 2026-07-14
fix: remove default value from resource node template

nkotikov-castai:
This PR removes default value from `spot_interruption_predictions` field and introduces a check on empty string.


castai/resource_node_template.go
```diff
@@ -1758,7 +1757,7 @@ func toTemplateConstraints(obj map[string]any) *sdk.NodetemplatesV1TemplateConst
 	if v, ok := obj[FieldNodeTemplateSpotInterruptionPredictionsEnabled].(bool); ok {
 		out.SpotInterruptionPredictionsEnabled = toPtr(v)
```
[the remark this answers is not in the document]
nkotikov-castai: Done

castai/resource_node_template.go
```diff
@@ -291,7 +291,6 @@ func resourceNodeTemplate() *schema.Resource {
 						FieldNodeTemplateSpotInterruptionPredictionsType: {
```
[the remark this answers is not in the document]
nkotikov-castai: >consider whether a state migration or DiffSuppressFunc is needed to avoid spurious diffs for existing resources that already have the default value in state

Do we really need that? It was previously not set.

primoz-hrovat (approved):

castai/resource_node_template.go
```diff
@@ -291,7 +291,6 @@ func resourceNodeTemplate() *schema.Resource {
...
 							Type:             schema.TypeString,
 							Optional:         true,
-							Default:          "interruption-predictions",
 							Description:      "Spot interruption predictions type. Only \"interruption-predictions\" is supported.",
 							Deprecated:       "The value \"aws-rebalance-recommendations\" is deprecated and will be removed in a future major version. Cast AI ML predictions (\"interruption-predictions\") are now used for all spot interruption prediction.",
 							ValidateDiagFunc: validation.ToDiagFunc(validation.StringInSlice([]string{"aws-rebalance-recommendations", "interruption-predictions"}, false)),
```
thomasstrang-cast, line 296: There was no default before. We added a default but the suppress function didn't cover the situation where a customer didn't have a default before but now does.

If we remove the default now then we might have the opposite problem, if a customer or two has now applied the default, then this MR will result in a new diff that removes the old default.

So we need to suppress the diff on this field to prevent the change to the default resulting in a diff. 
```suggestion
						Deprecated:       "The value \"aws-rebalance-recommendations\" is deprecated and will be removed in a future major version. Cast AI ML predictions (\"interruption-predictions\") are now used for all spot interruption prediction.",
						DiffSuppressFunc: func(k, oldValue, newValue string, d *schema.ResourceData) bool {
						    return true
						},
						ValidateDiagFunc: validation.ToDiagFunc(validation.StringInSlice([]string{"aws-rebalance-recommendations", "interruption-predictions"}, false)),
```

I expect I got the formatting of this suggestion wrong

thomasstrang-cast (changes requested):

castai/resource_node_template.go
nkotikov-castai: @thomasstrang-cast maybe better this? We'll surpress only those who populated default
```
							DiffSuppressFunc: func(k, oldVal, newVal string, d *schema.ResourceData) bool {
								return newVal == "" && oldVal == "interruption-predictions"
							},
```

thomasstrang-cast: OK Yes I am on the same page as you, that would work. I personally would just suppress everything as the field is deprecated, but I agree the above would work and is more specific. ^^

thomasstrang-cast (approved):

castai/resource_node_template.go
nkotikov-castai: @thomasstrang-cast kinda weird/interesting case: 
```
DiffSuppressFunc: func(k, oldVal, newVal string, d *schema.ResourceData) bool {
	return newVal == "" && oldVal == "interruption-predictions"
}
```

fails acceptance tests. However this fixes it: 
```
DiffSuppressFunc: func(k, oldVal, newVal string, d *schema.ResourceData) bool {
	return newVal == "" && oldVal == "interruption-predictions"
}
```

Also another approach passes: 
```
DiffSuppressFunc: func(k, oldVal, newVal string, d *schema.ResourceData) bool {
    if d.Id() == "" {
        return false // persist whatever the config resolves to
    }
    return true // never diff it
 }
```

thomasstrang-cast: Sounds good. 

I couldn't [find the difference](https://i.redd.it/6lwrp2xhplg41.jpg) between those first 2 sorry :D

nkotikov-castai: Oh sorry, the first one supposed to be `return true`

PR #760, opened 2026-07-28
fix: handle removing individual attributes from edge configuration

andrejatcastai:


castai/resource_edge_configuration.go
```diff
@@ -426,7 +430,16 @@ func (r *edgeConfigurationResource) Update(ctx context.Context, req resource.Upd
 		Cri:            criConfig,
 	}
 
-	apiResp, err := client.EdgeConfigurationsAPIUpdateEdgeConfigurationWithResponse(ctx, organizationID, clusterID, edgeLocationID, plan.ID.ValueString(), nil, updateReq)
+	updateMask := r.computeUpdateMask(plan, priorState)
```
varnastadeus: updateMask is something that is filled during runtime based on the request, I dont think we should populate it manually

andrejatcastai: @varnastadeus, yes, it is populated automatically now and it is causing a problem. If field is removed (for example in case imageID had a value and with the new request this value is dropped to instruct using a new default image), then there will be no mask for imageID. As a result the old imageID will remain in the new version of Node Configuration.

varnastadeus (approved):

PR #763, opened 2026-07-29
fix: prefer apply_threshold_strategy over deprecated apply_threshold

anthonyvg9:
## Summary
- Remove `ConflictsWith` between `apply_threshold` and `apply_threshold_strategy` on workload scaling policy CPU/memory resources.
- When both are set, prefer `apply_threshold_strategy` (needed because `castai/eks-cluster` historically always sets deprecated `apply_threshold = 0.1`, which blocked Dynamic/`DEFAULT_ADAPTIVE` configs).
- Existing single-field and neither-set behavior is unchanged (default remains Percentage 0.1).

## Test plan
- [x] Unit tests: `Test_resolveApplyThresholdStrategy` (neither / threshold-only / strategy-only / both)
- [x] Local provider validate: both fields set succeeds (no conflict)
- [x] Registry provider: both fields set still fails with conflict (baseline)
- [x] Live plan against cluster: both set → strategy wins; neither → falls back to Percentage 0.1
- [x] CI unit tests pass

### DIFF
--- a/castai/resource_workload_scaling_policy.go
+++ b/castai/resource_workload_scaling_policy.go
@@ -510,12 +510,12 @@ func workloadScalingPolicyResourceSchema(resource, function string, overhead, mi
 				Optional: true,
 				MaxItems: 1,
 				Description: "Resource apply threshold strategy settings. " +
-					"The default strategy is `PERCENTAGE` with percentage value set to 0.1.",
+					"The default strategy is `PERCENTAGE` with percentage value set to 0.1. " +
+					"When both apply_threshold and apply_threshold_strategy are set, apply_threshold_strategy takes precedence.",
 				Elem: workloadScalingPolicyResourceApplyThresholdStrategySchema(),
 				DiffSuppressFunc: func(k, old, new string, d *schema.ResourceData) bool {
 					return suppressThresholdStrategyDefaultValueDiff(resource, old, new, d)
 				},
-				ConflictsWith: []string{fmt.Sprintf("%s.0.%s", resource, DeprecatedFieldApplyThreshold)},
 			},
 			"look_back_period_seconds": {
 				Type:             schema.TypeInt,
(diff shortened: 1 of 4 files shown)


anthonyvg9:
@getkimchi review

anthonyvg9:
## Note for reviewers: unrelated CI failures

The following CI failures are **not caused by this PR** (WOOP `apply_threshold` / strategy preference). The same jobs are currently red on other open PRs as well:

1. **Terraform Examples Validation** — CI resolves `azurerm` to **v5.0.0**, which now requires `node_provisioning_profile` on `azurerm_kubernetes_cluster`. Our AKS examples do not set that block yet.
2. **Acceptance tests (`TestAccAKS_ResourceAKSCluster`)** — step 6/6 sets `aks.pod_subnet_id`; the API now rejects it for the shared test cluster CNI mode (`Kubenet` / Azure CNI Overlay are incompatible with `podSubnetId`).

Relevant green checks for this change: Go unit tests, golangci-lint, Terraform format, Generate SDK and Docs.

Suggested follow-up (separate infra PR): pin/constrain `azurerm` in examples and/or update AKS examples + acc test for the current CNI / azurerm v5 requirements.

Companion module PRs (omit `apply_threshold` when strategy is set; forward `only_if_original_*`):
- EKS: https://github.com/castai/terraform-castai-eks-cluster/pull/179
- GKE / AKS: sibling module PRs with the same change

castai/resource_workload_scaling_policy.go
```diff
@@ -510,12 +510,12 @@ func workloadScalingPolicyResourceSchema(resource, function string, overhead, mi
...
 				MaxItems: 1,
 				Description: "Resource apply threshold strategy settings. " +
-					"The default strategy is `PERCENTAGE` with percentage value set to 0.1.",
+					"The default strategy is `PERCENTAGE` with percentage value set to 0.1. " +
+					"When both apply_threshold and apply_threshold_strategy are set, apply_threshold_strategy takes precedence.",
 				Elem: workloadScalingPolicyResourceApplyThresholdStrategySchema(),
 				DiffSuppressFunc: func(k, old, new string, d *schema.ResourceData) bool {
 					return suppressThresholdStrategyDefaultValueDiff(resource, old, new, d)
 				},
-				ConflictsWith: []string{fmt.Sprintf("%s.0.%s", resource, DeprecatedFieldApplyThreshold)},
```
ZuzanaLietavcova, line 518: Allowing both fields causes perpetual diff in the plan - as the TF state contains only the new FieldApplyThresholdStrategy (also API drops the legacy one if we send both). While on the TF plan we would need to do some custom diff suppression.

PR #771, opened 2026-07-31
feat: add castai_commitment resource for generic commitments import

michal-kopczynski:


michal-kopczynski:
@kimchi review

kodburn (approved):
Do we only support GCP CUDs import here? Can AWS be added as well?

PR #774, opened 2026-08-04
feat: Add hpa converter support

mszostok:


castai/resource_workload_scaling_policy.go
```diff
@@ -1810,6 +1835,43 @@ func toAntiAffinityMap(s *sdk.WorkloadoptimizationV1AntiAffinitySettings) []map[
...
+	}
+	if len(out) == 0 {
+		return nil
+	}
+	return &out
+}
+
+func toHpaConvertersMap(s *[]sdk.WorkloadoptimizationV1HPAConverters) []map[string]any {
+	if s == nil {
+		return nil
```
ZuzanaLietavcova, line 1864: nit: could be nil for empty slice too

ZuzanaLietavcova (approved):

castai/resource_workload_scaling_policy_test.go
```diff
@@ -1481,6 +1489,101 @@ func Test_toJvmMap(t *testing.T) {
...
+			},
+		},
+		"should return multiple hpa converters": {
+			args: []any{
+				map[string]any{FieldHpaConverterType: FieldHpaConverterTypeAverageValueFromOriginalRequests},
+				map[string]any{FieldHpaConverterType: FieldHpaConverterTypeAverageValueFromOriginalRequests},
+			},
+			exp: &[]sdk.WorkloadoptimizationV1HPAConverters{
+				{Type: sdk.WorkloadoptimizationV1HPAConverterType(FieldHpaConverterTypeAverageValueFromOriginalRequests)},
+				{Type: sdk.WorkloadoptimizationV1HPAConverterType(FieldHpaConverterTypeAverageValueFromOriginalRequests)},
```
ZuzanaLietavcova, line 1518: Q outside of TF: We allow multiple HPA converters in the API, but creating CRD with >1 of them currently fails, right? https://gitlab.com/castai/kubecast/-/blob/master/lib/woop-crd/api/v1/recommendation_types.go#L631

mszostok: yes, and actually need to protect the API side too, but both API/TF and CRD syntax is prepared for the future where we will add more converters, now allowing more doesn't make sense. Once we will add more then we will remove the restriction of max 1 elem but syntax will stay the same, so no breaking change but more of relaxing rules. Thanks for catching that 🙏

PR #777, opened 2026-08-06
feat: add support for the two phase recommendations

mszostok:


castai/resource_workload_scaling_policy.go
```diff
@@ -1693,9 +1734,42 @@ func toStartup(startup map[string]any) *sdk.WorkloadoptimizationV1StartupSetting
 		result.PeriodSeconds = lo.ToPtr(int32(v))
```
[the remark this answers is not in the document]
mszostok: it's pre-existing code.

examples/resources/castai_workload_scaling_policy/resource.tf
```diff
@@ -55,6 +55,13 @@ resource "castai_workload_scaling_policy" "services" {
   }
   startup {
     period_seconds = 240
+    two_phase_recommendations {
+      enabled = true
```
mszostok, line 64: `two_phase_recommendations` enabled without upper `period_seconds` > 0 are no-op. Should we have some backend validation for that case @zrudzionis? as I can produce:
```
  startup {
    two_phase_recommendations {
      enabled = true
      requests_on_startup {
        cpu_cores  = 0.5
        memory_gib = 1.5
      }
    }
  }
```
but this will be just no-op, not sure if customer won't think that we detect startup automatically in such case 🤔

zrudzionis: This could be a breaking change? On paper it makes sense but do we want to couple these two features? I mean if someone configures 2phase recs and later want to set `period_seconds=0` they will fail to do so.

mszostok: it's true but dunno if I would treat it as a breaking change though 🤔 however adding it now can become tricky as there could be a cases (probably not) that someone already exploited this path and the `two_phase_recommendations` without period second 🤔 

so maybe I won't touch it :D thanks!

zrudzionis (approved):

PR #784, opened 2026-08-13
chore: is_default node template validation

laimonasr:


castai/resource_node_template.go
```diff
@@ -157,6 +157,29 @@ func (m nodeSelectorOperatorsSlice) Get(k sdk.K8sSelectorV1Operator) (string, bo
...
+func resourceNodeTemplateCustomizeDiff(_ context.Context, d *schema.ResourceDiff, _ any) error {
+	name := d.Get(FieldNodeTemplateName).(string)
+	isDefault, isDefaultSet := d.GetOk(FieldNodeTemplateIsDefault)
+	if !isDefaultSet {
+		return nil
+	}
+	return validateNodeTemplateIsDefault(name, isDefault.(bool))
+}
+
+func validateNodeTemplateIsDefault(name string, isDefault bool) error {
```
thomasstrang-cast, line 169: if we don't let the field be anything other that what we want it to be (i.e. true for the default-by-castai else false) then maybe we should just deprecate the field from the provider?

thomasstrang-cast (approved):

castai/resource_node_template.go
laimonasr: Yes I think we can do it

PR #787, opened 2026-08-17
CO-4293: integrate evictor and policies v2

sk91:


castai/sdk/cluster_autoscaler_v2/client.go
```diff
@@ -0,0 +1,25 @@
...
+		return nil
+	}
+
+	apiClient, err := NewClientWithResponses(apiURL, httpClientOption)
+	if err != nil {
+		return nil, err
+	}
+
+	return apiClient, nil
+}
```
jozuenoon: Should we have CreateClient as single common function for both autoscaler and evictor maybe as `pkg` or just under `sdk`? I takes same parameters even in `provider.go` later.

castai/resource_autoscaler_policies.go
```diff
@@ -0,0 +1,500 @@
...
+
+			out.PodPinner = podPinner
+		}
+	}
+
+	return out, nil
+}
+
+func setAutoscalerPoliciesState(data *schema.ResourceData, policies *cluster_autoscaler_v2.PoliciesV2) error {
+	if policies.Enabled != nil {
```
jozuenoon, line 390: I'm not sure if we should skip `data.Set`, this would most likely cause the state drift if something got changed manually in the cluster. Enabled would probably not ever be nil, but the other fields probably can be nil if unset. Just that `flatten*` functions should return nil if input was nil below.

castai/resource_eviction_config.go
```diff
@@ -168,13 +175,19 @@ func readAdvancedEvictorConfig(ctx context.Context, data *schema.ResourceData, m
...
 
-	resp, err := client.EvictorAPIGetAdvancedConfigWithResponse(ctx, clusterId)
+	resp, err := client.EvictorAPIGetEvictorAdvancedConfigWithResponse(ctx, clusterId)
 	if err != nil {
 		log.Printf("[ERROR] Failed to set read evictor advanced config: %v", err)
 		return err
 	}
+	if resp.StatusCode() != http.StatusOK {
+		return fmt.Errorf("expected status code %d, received: status=%d body=%s", http.StatusOK, resp.StatusCode(), string(resp.Body))
+	}
```
jozuenoon, line 192: We should handle 404, the API returns it while advanced config is nil.

jozuenoon:
```
if resp.StatusCode() == http.StatusNotFound {          // ← add this
        log.Printf("[INFO] Evictor advanced config for cluster %s not found, removing from state", clusterId)
        data.SetId("")
        return nil
    }
```

castai/resource_eviction_config.go
```diff
@@ -391,7 +425,7 @@ func toPodSelector(in interface{}) (*sdk.CastaiEvictorV1PodSelector, error) {
 	if !ok {
```
jozuenoon, line 430: by the way, the `ok` is missing in assertion?

castai/resource_eviction_config.go
```diff
@@ -468,7 +502,7 @@ func toNodeSelector(in interface{}) (*sdk.CastaiEvictorV1NodeSelector, error) {
...
```
jozuenoon, line 507: by the way, `ok` is missing in assertion

jozuenoon (approved):

PR #793, opened 2026-08-21
FIX - child_organization_id passing logic

oskarwojciski:
should be set only when organization is enterprise and scope organization id is different from enterprise_id

### DIFF
--- a/castai/resource_enterprise_role_binding.go
+++ b/castai/resource_enterprise_role_binding.go
@@ -470,8 +470,12 @@ func buildRoleBindingDefinition(roleBinding EnterpriseRoleBinding, enterpriseID 
 
 	// When the role binding targets a child organization, set ChildOrganizationId so the
 	// API validates role_id against the child org's roles instead of the enterprise's roles.
-	if roleBinding.OrganizationID != enterpriseID {
-		definition.ChildOrganizationId = lo.ToPtr(roleBinding.OrganizationID)
+	if roleBinding.OrganizationID == enterpriseID {
+		if len(scopes) > 0 {
+			if scopes[0].Organization != nil {
+				definition.ChildOrganizationId = lo.ToPtr(scopes[0].Organization.Id)
+			}
+		}
 	}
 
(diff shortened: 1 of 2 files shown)


castai/resource_enterprise_role_binding.go
```diff
@@ -470,8 +470,8 @@ func buildRoleBindingDefinition(roleBinding EnterpriseRoleBinding, enterpriseID
 
 	// When the role binding targets a child organization, set ChildOrganizationId so the
 	// API validates role_id against the child org's roles instead of the enterprise's roles.
-	if roleBinding.OrganizationID != enterpriseID {
-		definition.ChildOrganizationId = lo.ToPtr(roleBinding.OrganizationID)
+	if roleBinding.OrganizationID == enterpriseID && len(scopes) > 0 && scopes[0].Organization != nil {
```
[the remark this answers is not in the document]
oskarwojciski: multiple organization scopes are not supported

oanatmaria (approved):

castai/resource_enterprise_role_binding_test.go
```diff
@@ -129,8 +129,7 @@ func TestResourceEnterpriseRoleBindingCreateContext(t *testing.T) {
 						Name:        "engineering-viewer",
 						Description: lo.ToPtr("Engineering viewer role binding"),
 						Definition: organization_management.RoleBindingDefinition{
-							RoleId:              lo.ToPtr(roleID),
-							ChildOrganizationId: lo.ToPtr(organizationID1),
+							RoleId: lo.ToPtr(roleID),
```
plks, line 132:
```suggestion
							RoleId: new(roleID),
```
?

plks (approved):

castai/resource_enterprise_role_binding_test.go
oskarwojciski: We would need to migrate all occurrences of `lt.ToPtr`. I'll leave it for another PR

PR #801, opened 2026-09-03
feat: gatewayReplicas support for EdgeLocations

fra98:


castai/resource_edge_location.go
```diff
@@ -110,6 +111,17 @@ func (m *cniModel) Equal(other *cniModel) bool {
 	return m.Overlay.Equal(other.Overlay) && m.OverlayEncap.Equal(other.OverlayEncap)
 }
 
+type liqoModel struct {
```
claudiolor, line 114: Why do we need a separate model? Can't we put gatewayReplicas under networking?

claudiolor (approved):

castai/resource_edge_location.go
fra98: @claudiolor @anjmao cause it doesn't match the struct and hierarchy we have in EdgeCluster CRD and API itself. 
Under `liqo` we have liqo specific settings like override for LB address/ports), but they are not yet on TF, so this gatewayReplicas is the first introduced. 
While `networking` is for non-liqo generic settings like tunneledCIDRs and edge CNI

anjmao (approved):

PR #803, opened 2026-09-10
feat: edge configuration added for Nebius

michalwronski-castai:


castai/resource_edge_configuration.go
```diff
@@ -459,7 +476,14 @@ func (r *edgeConfigurationResource) Update(ctx context.Context, req resource.Upd
...
 
-	apiResp, err := client.EdgeConfigurationsAPIUpdateEdgeConfigurationWithResponse(ctx, organizationID, clusterID, edgeLocationID, plan.ID.ValueString(), nil, updateReq)
+	apiResp, err := client.EdgeConfigurationsAPIUpdateEdgeConfigurationWithResponse(
+		ctx,
+		organizationID, clusterID, edgeLocationID, plan.ID.ValueString(),
+		&omni.EdgeConfigurationsAPIUpdateEdgeConfigurationParams{
+			UpdateMask: lo.ToPtr("nebius.gpu_cluster,nebius.reservation_ids"),
+		},
+		updateReq,
+	)
```
fra98: Why is this needed for these two only? Shouldn't the mask be computed automatically by the API?

castai/data_source_edge_configuration.go
andrejatcastai, line 334: I think you missed to init those fields here as well ;)

andrejatcastai (approved):

fra98 (approved):

PR #808, opened 2026-09-16
Fix edge location zones

anjmao:
We can't use zones with typed model because in some places zones are assigned from data source. Data source is unknown at plan time and will fail.


anjmao:
Fix is not needed. We should avoid depends_on pattern

